PR #4, opened 2021-05-04
Upgrade Quarkus version to 2.0.0.Alpha1

Sgitario:


Sgitario:
It's failing to something related to testcontainers... Won't do for now.

PR #55, opened 2021-05-25
Ability to add test tracing

pjgg:
The test framework will trace all your test/method invocations, so you can review how much time took to run a test
or filter by tags or errors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,6 +40,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <surefire-plugin.version>2.22.2</surefire-plugin.version>
         <testcontainers.version>1.15.3</testcontainers.version>
+        <jaeger-client.version>1.6.0</jaeger-client.version>
         <maven-checkstyle-plugin.version>3.1.2</maven-checkstyle-plugin.version>
         <formatter-maven-plugin.version>2.15.0</formatter-maven-plugin.version>
         <impsort-maven-plugin.version>1.6.2</impsort-maven-plugin.version>
@@ -131,6 +132,11 @@
                 <groupId>io.quarkus.qe</groupId>
                 <artifactId>quarkus-test-service-jaeger</artifactId>
                 <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.jaegertracing</groupId>
+                <artifactId>jaeger-client</artifactId>
+                <version>${jaeger-client.version}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
@@ -342,6 +348,7 @@
                 <module>examples/jaeger</module>
                 <module>examples/database-mysql</module>
                 <module>examples/external-applications</module>
+                <module>examples/tracing-extension</module>
             </modules>
         </profile>
         <profile>
--- a/quarkus-test-core/pom.xml
+++ b/quarkus-test-core/pom.xml
@@ -39,5 +39,9 @@
             <groupId>net.sourceforge.htmlunit</groupId>
             <artifactId>htmlunit</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.jaegertracing</groupId>
+            <artifactId>jaeger-client</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/QuarkusScenarioBootstrap.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/QuarkusScenarioBootstrap.java
@@ -25,13 +25,14 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.LifecycleMethodExecutionExceptionHandler;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolver;
+import org.junit.jupiter.api.extension.TestWatcher;
 
 import io.quarkus.test.logging.Log;
 import io.quarkus.test.utils.FileUtils;
 
 public class QuarkusScenarioBootstrap
-        implements BeforeAllCallback, AfterAllCallback, BeforeEachCallback, ParameterResolver,
-        LifecycleMethodExecutionExceptionHandler {
+        implements BeforeAllCallback, AfterAllCallback, BeforeEachCallback,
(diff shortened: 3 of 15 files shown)


quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/QuarkusScenarioBootstrap.java
```diff
@@ -98,18 +99,46 @@ public void handleAfterAllMethodExecutionException(ExtensionContext context, Thr
...
     @Override
     public void handleBeforeEachMethodExecutionException(ExtensionContext context, Throwable throwable) {
         notifyExtensionsOnError(context, throwable);
     }
 
+    private void launchServices(ExtensionContext context, Service service) {
+        try {
+            testFrameworkTracer.createSpanContext(context);
+            service.start();
+            testFrameworkTracer.finishWithSuccess(context, Sets.newHashSet(service.getName()));
```
Sgitario, line 134: I don't see the value of adding the service name as a tag :/

quarkus-test-core/src/main/java/io/quarkus/test/tracing/TestFrameworkTracer.java
```diff
@@ -0,0 +1,73 @@
...
+
+import io.jaegertracing.internal.JaegerTracer;
+import io.jaegertracing.internal.reporters.RemoteReporter;
+import io.jaegertracing.internal.samplers.ConstSampler;
+import io.jaegertracing.thrift.internal.senders.HttpSender;
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+import io.opentracing.log.Fields;
+
+public class TestFrameworkTracer {
```
Sgitario, line 22: I would rename the class to QuarkusScenarioTracer.

quarkus-test-core/src/main/java/io/quarkus/test/tracing/TestFrameworkTags.java
```diff
@@ -0,0 +1,34 @@
+package io.quarkus.test.tracing;
+
+import java.util.Map;
+import java.util.Set;
+
+import io.opentracing.Span;
+import io.opentracing.tag.Tags;
+
+public class TestFrameworkTags {
```
Sgitario, line 9: Why this logic can't go in the tracer?

quarkus-test-core/src/main/java/io/quarkus/test/tracing/TestFrameworkTracer.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private static final String DEFAULT_SERVICE_NAME = "quarkus-test-framework";
+    private static final String DEFAULT_JAEGER_HTTP_ENDPOINT = "http://localhost:14268/api/traces";
```
Sgitario, line 25: Use `PropertyLookup` for this purpose that allows you to set a defaults.
Also, all the properties for the test framework should start with "ts."

Sgitario (changes requested):
Summary:
- I think the tracer should be an optional feature (it should implement `ExtensionBootstrap`).
The `QuarkusScenarioBootstrap` should do only a single thing: bootstrap the test components (services, extensions, annotations).
- Properties should start with "ts."
- Missing coverage of this feature (I think we could add some test development to ensure the correct integration with Jaeger).

quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/QuarkusScenarioBootstrap.java
pjgg: In the end is easier to filter by Jaeger

quarkus-test-core/src/main/java/io/quarkus/test/tracing/TestFrameworkTags.java
pjgg: Because of Tracer, Tags, Span... start growing in terms of methods and I think that is easier to maintain with segregated responsibilities.

pjgg:
Missing coverage of this feature

examples/TracingExtension/pom.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.qe</groupId>
+        <artifactId>quarkus-test-parent</artifactId>
+        <version>0.0.2-SNAPSHOT</version>
+        <relativePath>../../</relativePath>
+    </parent>
+    <artifactId>examples-tracingExtension</artifactId>
```
Sgitario, line 10: In order to follow the rest of the examples name convention, it should be: `examples-tracing-extension` or maybe simply `examples-tracing`.

examples/TracingExtension/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+    <name>Quarkus - Test Framework - Examples - tracingExtension</name>
```
Sgitario, line 11: Same here: Quarkus - Test Framework - Example - Tracing

pom.xml
```diff
@@ -342,6 +342,7 @@
                 <module>examples/jaeger</module>
                 <module>examples/database-mysql</module>
                 <module>examples/external-applications</module>
+                <module>examples/TracingExtension</module>
```
Sgitario, line 345: same: tracing or tracing-extension

quarkus-test-core/pom.xml
```diff
@@ -39,5 +39,9 @@
             <groupId>net.sourceforge.htmlunit</groupId>
             <artifactId>htmlunit</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-opentracing</artifactId>
```
Sgitario, line 44: Do we need this dependency? This is bound to Quarkus releases and as far I see, we only need it for `io.jaegertracing` and `o.opentracing`, so we could add the jaeger dependencies directly. This way we keep the dependencies with Quarkus as lower as possible.

Sgitario (approved):

PR #60, opened 2021-05-27
Support Quarkus CLI testing

Sgitario:
The Quarkus Test Framework supports the usage of [the Quarkus CLI tool](https://quarkus.io/version/main/guides/cli-tooling):

```java
@QuarkusScenario
public class QuarkusCliClientIT {

    @Inject
    static QuarkusCliClient cliClient;

    @Test
    public void shouldVersionMatchQuarkusVersion() {
        String cliVersion = cliClient.getVersion();
        assertEquals("Client Version " + Version.getVersion(), cliVersion);
    }

    @Test
    public void shouldCreateApplicationOnJvm() {
        // Create application
        QuarkusCliRestService app = cliClient.createApplication("app");

        // Should build on Jvm
        QuarkusCliClient.Result result = app.buildOnJvm();
        assertTrue(result.isSuccessful(), "The application didn't build on JVM. Output: " + result.getOutput());

        // Start using DEV mode
        app.start();
        app.given().get().then().statusCode(HttpStatus.SC_OK);
    }
}
```

Current features:
- `run` - run any command
- `createApplication` - create a service at `target/<APP NAME>`
- `buildOnJvm` - build the service in JVM mode
- `buildOnNative` - build the service in Native mode
- `runOnDev` - run the service on DEV mode (it's the same as using `QuarkusCliRestService.start`)
- `getInstalledExtensions` - get the installed extensions
- `installExtension` - install a concrete Quarkus extension
- `removeExtension` - remove a concrete Quarkus extension

The framework will not install the Quarkus CLI tool, so before running these scenarios, it needs to be installed it beforehand.
The default command name is `quarkus`, but it can be changed using the property `ts.quarkus.cli.cmd`. For example:

```
mvn clean verify -Dts.quarkus.cli.cmd="java -jar quarkus-cli.jar"
```

The above command will use directly the binary from Quarkus upstream build.


quarkus-test-core/src/main/java/io/quarkus/test/utils/ProcessUtils.java
```diff
@@ -0,0 +1,32 @@
+package io.quarkus.test.utils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class ProcessUtils {
```
rsvoboda, line 7: We need support for Windows too.

Sgitario: To be honest, I don't like to have to do things like this... 
If we were using directly the CLI (instead of directly the Maven dependency - the ide Maven dependency), a lot of things would be much more easier ... Indeed, I wanted to discuss this with you the next week.

Sgitario:
run tests

Sgitario:
@rsvoboda this PR is ready to be reviewed now.

rsvoboda:
How much time do I have :) ?

rsvoboda:
I miss CLI example in `examples` directory, can it be added?

Sgitario:
> I miss CLI example in `examples` directory, can it be added?

I've just moved the existing test from the cli module to a new examples cli module.

quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/EnabledOnNativeCondition.java
```diff
@@ -0,0 +1,19 @@
...
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import io.quarkus.test.services.quarkus.model.QuarkusProperties;
+
+public class EnabledOnNativeCondition implements ExecutionCondition {
+
+    @Override
+    public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
```
rsvoboda, line 12: I don't find these messages intuitive, when I saw them for the first time in test log I got confused.

quarkus-test-core/src/main/java/io/quarkus/test/utils/FileUtils.java
```diff
@@ -37,11 +36,9 @@ public static Path copyContentTo(String content, Path target) {
 
     public static String loadFile(File file) {
         try {
-            return IOUtils.toString(
-                    new FileInputStream(file),
-                    StandardCharsets.UTF_8);
+            return org.apache.commons.io.FileUtils.readFileToString(file, StandardCharsets.UTF_8);
```
rsvoboda, line 39: Why commons util class instead of `String content = new String(Files.readAllBytes(Paths.get(fileName)));` ?

Sgitario: Just to have a single point where users can find file utilities for the test framework.
Still, if it causes confusion in the future, we can get rid of it. However, I don't want to change anything about this as it's unrelated to this PR.

quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/EnabledOnNativeCondition.java
Sgitario: I totally agree with you. I will change it.

quarkus-test-core/src/main/java/io/quarkus/test/utils/FileUtils.java
rsvoboda: My comment is for usage of `org.apache.commons.io.FileUtils.readFileToString`, not about `quarkus-test-core/src/main/java/io/quarkus/test/utils/FileUtils.java` 

I think we can leave this as is

quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/EnabledOnNativeCondition.java
Sgitario: Changed, let me know it's clearer now.
Thanks!

rsvoboda (approved):

rsvoboda:
Will wait for CI before merging.

Issue #72, opened 2021-05-31
Configuration: Support global properties for services


Sgitario:
Support configuring properties for all services using a `global.properties`, so we can configure properties for all the services at once:

```
ts.global.log.enable=true
```

Will enable logging for all the services.

PR #73, opened 2021-05-31
Support global properties. Log is enabled by default now.

Sgitario:
We now support configuring properties for all services using a `global.properties`, so we can configure properties for all the services at once:

```
ts.global.log.enable=true
```

Will enable logging for all the services.

Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/71
Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/72

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,26 @@ public class PingPongResourceIT {
 
 As seen in the above example, everything is bounded to a Service object that will contain everything needed to interact with our resources.
 
+### Configuration
+
+Test framework allows to customise the configuration for running the test case via a `test.properties` file placed under `src/test/resources` folder.
+
+The current configuration options are: 
+
+- Enable/Disable logging (enabled by default):
+
+```
+ts.<YOUR SERVICE NAME>.log.enable=true
+```
+
+Where `<YOUR SERVICE NAME>` could be either `pingApp`, `pongApp` or `pingPongApp` following the [Getting Started](#getting-started) example. If you want to use the same property for all your services, there is a special scope called `global` for such purposes:
+
+```
+ts.global.log.enable=false
+```
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,6 +51,26 @@ public class PingPongResourceIT {
 
 As seen in the above example, everything is bounded to a Service object that will contain everything needed to interact with our resources.
 
+### Configuration
+
+Test modules allow to customise the configuration for running the test case via a `test.properties` file.
```
kshpak, line 56:
```suggestion
Test modules allowed to customize the configuration for running the test case via a `test.properties` file placed under the `/resource` folder.
```

Sgitario: Changed

README.md
```diff
@@ -51,6 +51,26 @@ public class PingPongResourceIT {
...
+
+The current configuration options are: 
+
+- Enable/Disable logging (enabled by default):
+
+```
+ts.<YOUR SERVICE NAME>.log.enable=true
+```
+
+Where `<YOUR SERVICE NAME>` is the scope of the configuration property. If you want to use the same property for all your services, there is a special scope called `global` for such purposes:
```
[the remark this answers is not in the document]
Sgitario: Added some clarification around this.

README.md
```diff
@@ -51,6 +51,26 @@ public class PingPongResourceIT {
...
+
+The current configuration options are: 
+
+- Enable/Disable logging (enabled by default):
+
+```
+ts.<YOUR SERVICE NAME>.log.enable=true
+```
+
+Where `<YOUR SERVICE NAME>` could be either `pingApp`, `pongApp` or `pingPongApp` following the Getting Started example. If you want to use the same property for all your services, there is a special scope called `global` for such purposes:
```
kshpak, line 66: place Getting Started as a hyperlink to a chapter

Sgitario: Done

Issue #70, opened 2021-05-31
CI: Ensure OpenShift coverage


Sgitario:

Sgitario:
The CI coverage will be done for OpenShift only for now as secrets can't be shared across different forks in GitHub (k8s).

PR #74, opened 2021-05-31
Build JVM + Native + Kubernetes in the PR workflow

Sgitario:
Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/70
The OpenShift CI is already configured too.


Sgitario:
Secrets can't be shared across different forks. Closing.

Issue #75, opened 2021-05-31
Configuration: Allow to setup default factor timeouts


Sgitario:
Allow configuring timeouts for services. For example, in order to configure startup service timeout, we could use:

```
ts.<SERVICE_NAME>.timeout.startup=10 # 10 minutes
```

The default for startup is 5 minutes. 

Moreover, let's allow to have some factor times to adapt our test suite to slower environments:

```
ts.global.timeout.factor=2.5 # Default 1.
```

This value will be used to multiply the timeouts (timeout x factor = total timeout). Example: if we define a timeout of 5 minutes with a factor of 2.5, the total timeout will be 5 x 2.5 = 12.5 minutes.

The factor should be suitable to be supplied from system properties.

At the moment, let's support only the startup timeout.

rsvoboda:
`ts.<SERVICE_NAME>.timeout.startup=10m` looks more natural and there could be variant for seconds, e.g. `30s`

Sgitario:
> `ts.<SERVICE_NAME>.timeout.startup=10m` looks more natural and there could be variant for seconds, e.g. `30s`

Do you mean that you prefer to deal with timeouts without another `factor` property? Or just that you prefer to define timeout using duration annotation?

rsvoboda:
`ts.<SERVICE_NAME>.timeout.startup=10` = > `ts.<SERVICE_NAME>.timeout.startup=10m`

to make it self describing what unit is used

PR #77, opened 2021-05-31
Allow to configure timeouts via system properties / configuration files

Sgitario:
### Configuration

Test framework allows to customise the configuration for running the test case via a `test.properties` file placed under `src/test/resources` folder.

All the properties can be configured globally by replacing `<YOUR SERVICE NAME>` with `global`.

- Timeouts

Timeouts are quite important property to, as an example, control how long to wait for a service to start. The existing options to configure timeouts are::
 
```
# Startup timeout for services is 5 minutes
ts.<YOUR SERVICE NAME>.startup.timeout=5m
# Install operator timeout is 10 minutes
ts.<YOUR SERVICE NAME>.operator.install.timeout=10m
# Install image stream timeout is 5 minutes
ts.<YOUR SERVICE NAME>.imagestream.install.timeout=5m
```

For example, we can use `ts.pingApp.startup.timeout` to wait up to 10 min for the ping application to start. 

How often does the test framework check for startup conditions? This is also configurable via:

```
ts.<YOUR SERVICE NAME>.startup.check-poll-interval=2s
```

Now, we're saying that we need to wait up to 10 minutes for our service to start and will be checking the condition every 2 seconds.

How can we manipulate the overall timeout in slower environments? Let's say that our environment is twice and a half slower than a good environment, then we can instruct the test framework with:

```
ts.global.factor.timeout=2.5
```

And all the timeouts will take into account this factor value. For example, if previously, the startup timeout was 10 minutes, now it will be 10 * 2.5 = 25 minutes.

Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/75

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ As seen in the above example, everything is bounded to a Service object that wil
 
 Test framework allows to customise the configuration for running the test case via a `test.properties` file placed under `src/test/resources` folder.
 
+All the properties can be configured globally by replacing `<YOUR SERVICE NAME>` with `global`.
+
 The current configuration options are: 
 
 - Enable/Disable logging (enabled by default):
@@ -70,6 +72,37 @@ ts.global.log.enable=false
 ```
 
 The above configuration will disable logging for all your services. The same can be set via system properties by running `-Dts.global.log.enable=false`.
+
+- Timeouts
+
+Timeouts are quite important property to, as an example, control how long to wait for a service to start. The existing options to configure timeouts are:
+ 
+```
+# Startup timeout for services is 5 minutes
+ts.<global or YOUR SERVICE NAME>.startup.timeout=5m
+# Default startup check poll interval is every 2 seconds
+ts.<global or YOUR SERVICE NAME>.startup.check-poll-interval=2s
+# Install operator timeout is 10 minutes
+ts.<global or YOUR SERVICE NAME>.operator.install.timeout=10m
+# Install image stream timeout is 5 minutes
+ts.<global or YOUR SERVICE NAME>.imagestream.install.timeout=5m
+```
+
+In order to increase the default timeout for all the services, we can use the `global` scope. For example, to increase the default startup timeout: `ts.global.startup.timeout=10m`. Also, we can configure how often the test framework will check for the startup condition using the property `ts.global.startup.check-poll-interval=3s`. Using these two properties, we are making all the services to wait up to 10 minutes to start and checking the condition every 3 seconds.
+
+On the other hand, if we want to update the timeout for a single service because we know that this service is quite slow, the scope of the property is the service name. For example, let's imagine that the `pingApp` service from the [Getting Started](#getting-started) section is very slow, we can use `ts.pingApp.startup.timeout=20m` to wait up to 20 min for only the ping application to start. 
+
+How can we manipulate the overall timeout in slower environments? Let's say that our environment is twice and a half slower than a good environment, then we can instruct the test framework with:
+
+```
+ts.global.factor.timeout=2.5
+```
+
+And all the timeouts will take into account this factor value. For example, if previously, the startup timeout was 10 minutes, now it will be 10 * 2.5 = 25 minutes.
+
+In a Multi-Module Maven test suite, if we want to configure the timeouts, we can do this via system properties:
+- Increase the startup timeout only: `mvn clean verify -Dts.global.startup.timeout=20m`
+- Increase all the timeouts at once using the factor: `mvn clean verify -Dts.global.factor.timeout=2.5`
 
 ### Native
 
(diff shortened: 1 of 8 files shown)


rsvoboda (commented):
Let's assume  multi module project
```
.
├── module-001
├── module-002
└── module-003
```
Can I have one global timeout settings file which would be applicable to all the 3 modules?

Sgitario:
> Let's assume multi module project
> 
> ```
> .
> ├── module-001
> ├── module-002
> └── module-003
> ```
> 
> Can I have one global timeout settings file which would be applicable to all the 3 modules?

Yes, we have two ways to do so:

- Update a single type of timeout, for example, the timeout to wait for a service to start:

```
ts.global.startup.timeout=20m
```

- Or update all the timeout at once using the factor timeout:

```
ts.global.factor.timeout=2.5
```

It will multiply all the timeouts by 2.5. (the default value is 1).

quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/BaseService.java
```diff
@@ -22,8 +22,10 @@
 
 public class BaseService<T extends Service> implements Service {
 
-    private static final int SERVICE_WAITER_POLL_EVERY_SECONDS = 2;
-    private static final int SERVICE_WAITER_TIMEOUT_MINUTES = 5;
+    private static final String SERVICE_STARTUP_CHECK_POLL_INTERVAL = "startup.check-poll-interval";
+    private static final Duration SERVICE_STARTUP_CHECK_POLL_INTERVAL_DEFAULT = Duration.ofSeconds(2);
```
rsvoboda, line 26: default poll and startup timeouts should be mentioned in README

rsvoboda:
Main question is where to put the property file with `ts.global.startup.timeout=20m` to be picked for all the 3 modules ?

quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/BaseService.java
Sgitario: It's in: https://github.com/quarkus-qe/quarkus-test-framework/pull/77/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R82 and https://github.com/quarkus-qe/quarkus-test-framework/pull/77/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R94

Sgitario:
> Main question is where to put the property file with `ts.global.startup.timeout=20m` to be picked for all the 3 modules ?



> Main question is where to put the property file with `ts.global.startup.timeout=20m` to be picked for all the 3 modules ?

The easiest way is via system property: `mvn clean verify -Dts.global.startup.timeout=20m`. Another way is via having a global config file to these settings. At the moment, the defaults are defined in https://github.com/quarkus-qe/quarkus-test-framework/blob/79710b037159e8a226099291b095fdfaa14d6c88/quarkus-test-core/src/main/resources/global.properties.

rsvoboda:
OK, so the test framework defines some default that would be used in any project that uses our framework.

README should be enhance with the content similar to your last comment @Sgitario

Sgitario:
> OK, so the test framework defines some default that would be used in any project that uses our framework.
> 
> README should be enhance with the content similar to your last comment @Sgitario

Ok, I've already added it. New lines are 107 to 109: https://github.com/quarkus-qe/quarkus-test-framework/pull/77/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R107

README.md
```diff
@@ -71,6 +73,41 @@ ts.global.log.enable=false
 
 The above configuration will disable logging for all your services. The same can be set via system properties by running `-Dts.global.log.enable=false`.
 
+- Timeouts
+
+Timeouts are quite important property to, as an example, control how long to wait for a service to start. The existing options to configure timeouts are::
```
rsvoboda, line 78:
```suggestion
Timeouts are quite important property to, as an example, control how long to wait for a service to start. The existing options to configure timeouts are:
```

quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/BaseService.java
rsvoboda: I think they should be highlighted more

rsvoboda: Maybe link quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/BaseService.java ?

`Consult quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/BaseService.java and quarkus-test-core/src/main/resources/global.properties to see the defaults` wdyt?

Sgitario: I've updated the README by trying to make it clearer. 
I didn't add any link but let me know if you still prefer adding it.

rsvoboda (approved):

Issue #78, opened 2021-06-01
Dev Mode ++ OpenShift extension: Test execution fails when test module requires parent dependencies


Sgitario:
When using [DEV mode](https://github.com/quarkus-qe/quarkus-test-framework#dev-mode) or other Quarkus OpenShift extensions to deploy the Quarkus application, the test framework is internally copying the Maven pom.xml of the example and running an inner Maven command.
The problem of this approach is when the pom.xml of the example uses a parent pom dependency that has not been previously installed.
There are different approaches to fix this issue:
1. Document that the framework needs to install the parents beforehand: This is a really poor intuitive
2. Obey to not have parent dependencies for examples: Again this is a poor limitation we should avoid
3. Automatically install the parent poms before running the inner Maven command.

Sgitario:
I implemented the third option here: https://github.com/quarkus-qe/quarkus-test-framework/pull/79
However, if you see other approaches, I'm very welcome.
cc @kshpak

kshpak:
@Sgitario I like your solution, but I have the following questions:
1. Does framework copy testcase directory into some temporary directory?

I mean, if we need to install parent pom first, must be the testcase is somehow separated from the testsuite

Sgitario:
> @Sgitario I like your solution, but I have the following questions:
> 
> 1. Does framework copy testcase directory into some temporary directory?
> 
> I mean, if we need to install parent pom first, must be the testcase is somehow separated from the testsuite

Yes, but the main problem is that we're running the test case module independently and hence needing to have all the dependencies resolved beforehand.

PR #79, opened 2021-06-01
Automatically install Parent Poms when using DevMode and OpenShift

Sgitario:
With these changes, the framework will ensure that the inner Maven command will work even when the test module uses a parent dependency.

Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/78


kshpak (approved):

kshpak:
It works, code is easy to understand. @Sgitario I will leave merging to you, maybe you will come up with some enhancement in the meanwhile

PR #81, opened 2021-06-01
Add Tracing information and Jaeger openshift template for installation

Sgitario:
Moved Tracing documentation into a separated document.
Added Jaeger installation template.


Sgitario:
Merging this PR as it's only for documentation.

Issue #76, opened 2021-05-31
Logging: fix logs output alignment


kshpak:
Logs lines are broken into pieces which makes it hard to read
![image](https://user-images.githubusercontent.com/71509020/120186004-325c2980-c213-11eb-8ed0-5b88ed4baa47.png)

PR #83, opened 2021-06-02
Fix logs printing for localhost

kshpak:
Fix for https://github.com/quarkus-qe/quarkus-test-framework/issues/76

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/logging/FileQuarkusApplicationLoggingHandler.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/logging/FileQuarkusApplicationLoggingHandler.java
@@ -1,38 +1,24 @@
 package io.quarkus.test.logging;
 
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.nio.file.Path;
+import java.io.File;
 
 import io.quarkus.test.bootstrap.ServiceContext;
+import io.quarkus.test.utils.FileUtils;
 
 public class FileQuarkusApplicationLoggingHandler extends LoggingHandler {
 
-    private static final int FILE_BUFFER_SIZE = 8 * 1024;
+    private final File file;
+    private String printedContent;
 
-    private final Path file;
-    private final InputStream from;
-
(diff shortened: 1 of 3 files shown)


quarkus-test-core/src/main/java/io/quarkus/test/logging/FileQuarkusApplicationLoggingHandler.java
```diff
@@ -1,38 +1,32 @@
...
 public class FileQuarkusApplicationLoggingHandler extends LoggingHandler {
 
     private static final int FILE_BUFFER_SIZE = 8 * 1024;
 
-    private final Path file;
-    private final InputStream from;
+    private final File file;
 
-    public FileQuarkusApplicationLoggingHandler(ServiceContext context, String output, InputStream from) {
+    public FileQuarkusApplicationLoggingHandler(ServiceContext context, File output) {
```
Sgitario, line 16: Rename output to `from` or `input`.

quarkus-test-core/src/main/java/io/quarkus/test/logging/FileQuarkusApplicationLoggingHandler.java
```diff
@@ -1,38 +1,32 @@
...
         super(context);
-        this.file = context.getServiceFolder().resolve(output);
-        this.from = from;
+        this.file = output;
     }
 
     @Override
     protected void handle() {
```
Sgitario, line 22: This implementation is printing the whole file all the time. We need to print only the differences and this is indeed what the onStringDifference is doing, basically, the implementation should be:

```
String newContent = FileUtils.loadFile(from);
onStringDifference(newContent, oldContent);
oldContent = newContent;
```

Sgitario (changes requested):

quarkus-test-core/src/main/java/io/quarkus/test/logging/FileQuarkusApplicationLoggingHandler.java
kshpak: Didn't realize that. Changed

Sgitario (approved):

PR #84, opened 2021-06-03
Support of global properties location

Sgitario:


mjurc (approved):
Looks good to me, thanks for the PR!

PR #95, opened 2021-06-08
Support Custom Service Start Up order

Sgitario:
By default, the services are initialized in Natural Order of presence. For example, having:

```java
class MyParent {
    @QuarkusApplication // ... or @Container
    static final RestService firstAppInParent = new RestService();
    
    @QuarkusApplication // ... or @Container
    static final RestService secondAppInParent = new RestService();

}

@QuarkusScenario
class MyScenarioIT extends MyParent {
    @QuarkusApplication // ... or @Container
    static final RestService firstAppInChild = new RestService();

    @QuarkusApplication // ... or @Container
    static final RestService secondAppInChild = new RestService();
}
```

Then, the framework will initialize the services at this order: `firstAppInParent`,  `secondAppInParent`, `firstAppInChild` and `secondAppInChild`.

We can change this order by using the `@LookupService` annotation:

```java
class MyParent {
    @LookupService
    static final RestService appInChild;

    @QuarkusApplication // ... or @Container
    static final RestService appInParent = new RestService().withProperty("x", () -> appInChild.getHost());
}

@QuarkusScenario
class MyScenarioIT extends MyParent {
    @QuarkusApplication // ... or @Container
    static final RestService appInChild = new RestService().dependsOn(appInParent);
}
```

Now, the framework will initialize the `appInChild` service first and then the `appInParent` service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,54 @@ public class PingPongResourceIT {
 
 As seen in the above example, everything is bounded to a Service object that will contain everything needed to interact with our resources.
 
+### Services Start Up Order
+
+By default, the services are initialized in Natural Order of presence. For example, having:
+
+```java
+class MyParent {
+    @QuarkusApplication // ... or @Container
+    static final RestService firstAppInParent = new RestService();
+    
+    @QuarkusApplication // ... or @Container
+    static final RestService secondAppInParent = new RestService();
+
+}
+
+@QuarkusScenario
+class MyScenarioIT extends MyParent {
+    @QuarkusApplication // ... or @Container
+    static final RestService firstAppInChild = new RestService();
+
+    @QuarkusApplication // ... or @Container
+    static final RestService secondAppInChild = new RestService();
+}
+```
+
+Then, the framework will initialize the services at this order: `firstAppInParent`,  `secondAppInParent`, `firstAppInChild` and `secondAppInChild`.
+
+We can change this order by using the `@LookupService` annotation:
+
+```java
+class MyParent {
+    @LookupService
+    static final RestService appInChild; // field name must match with the service name declared in MyScenarioIT.
+
+    @QuarkusApplication // ... or @Container
+    static final RestService appInParent = new RestService().withProperty("x", () -> appInChild.getHost());
+}
+
+@QuarkusScenario
+class MyScenarioIT extends MyParent {
+    @QuarkusApplication // ... or @Container
+    static final RestService appInChild = new RestService();
+}
+```
+
+| Note that field name of the `@LookupService` must match with the service name declared in MyScenarioIT.
+
+Now, the framework will initialize the `appInChild` service first and then the `appInParent` service.
+
 ### Configuration
 
 Test framework allows to customise the configuration for running the test case via a `test.properties` file placed under `src/test/resources` folder.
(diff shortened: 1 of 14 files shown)


quarkus-test-core/src/main/java/io/quarkus/test/utils/ReflectionUtils.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public static List<Field> findAllFields(Class<?> clazz) {
+        if (clazz == Object.class) {
+            return Collections.emptyList();
+        }
+
+        List<Field> fields = new ArrayList<>();
+        fields.addAll(findAllFields(clazz.getSuperclass()));
+        fields.addAll(Arrays.asList(clazz.getDeclaredFields()));
```
jsmrcka, line 50: `clazz.getDeclaredFields()` does not guarantee an order. So we cannot be sure that the services from within a single class will be initialized in the order of presence.
The `@LookupService` can only be used in a parent-child setup, but does not address situations like this:
```java
@QuarkusScenario
class MyScenarioIT {
    @QuarkusApplication
    static final RestService appOne = new RestService();

    @QuarkusApplication
    static final RestService appTwo = new RestService().withProperty("x", () -> appOne.getHost());
}
```

README.md
```diff
@@ -82,6 +82,52 @@ public class PingPongResourceIT {
...
+    static final RestService appInChild;
+
+    @QuarkusApplication // ... or @Container
+    static final RestService appInParent = new RestService().withProperty("x", () -> appInChild.getHost());
+}
+
+@QuarkusScenario
+class MyScenarioIT extends MyParent {
+    @QuarkusApplication // ... or @Container
+    static final RestService appInChild = new RestService().dependsOn(appInParent);
```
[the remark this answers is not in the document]
Sgitario: Ops, this is a left over. Thanks for spotting!

README.md
```diff
@@ -82,6 +82,52 @@ public class PingPongResourceIT {
...
+}
+```
+
+Now, the framework will initialize the `appInChild` service first and then the `appInParent` service.
+
```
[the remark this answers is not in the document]
Sgitario: +100

quarkus-test-core/src/main/java/io/quarkus/test/utils/ReflectionUtils.java
Sgitario: This scenario is already supported. Services are init by order of presence within the same class.

About the `clazz.getDeclaredFields()`, the order is not guaranteed when all the fields are independent from each other. For example:

```
@QuarkusScenario
class MyScenarioIT {
    @QuarkusApplication
    static final RestService appOne = new RestService();

    @QuarkusApplication
    static final RestService appTwo = new RestService();
}
```

Here, it's true that the bytecode output is not guaranteing the field order, but here we don't really care as services do not depend on each other.

For this scenario:

```
@QuarkusScenario
class MyScenarioIT {
    @QuarkusApplication
    static final RestService appOne = new RestService();

    @QuarkusApplication
    static final RestService appTwo = new RestService().withProperty("x", () -> appOne.getHost());
}
```

The field order is guaranteed by the bytecode output (which is what the clazz.getDeclaredFields method is based on).

jsmrcka: You're right, I didn't realize that. Thanks for explaining. :)

jsmrcka (approved):

PR #105, opened 2021-06-14
Fix binding of ts.s2i.maven.remote.repository property and document it

Sgitario:
The property `ts.s2i.maven.remote.repository` has been renamed to `ts.global.s2i.maven.remote.repository`, so we can configure it by service scope (using `ts.myService.s2i.maven.remote.repository`).

Also, it has been fixed as before this property was not been replaced.
Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/66

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,6 +696,33 @@ public class OpenShiftS2iQuickstartIT {
     //
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -696,6 +696,33 @@ public class OpenShiftS2iQuickstartIT {
     //
 ```
 
+It's important to note that, by default, OpenShift will build the application's source code using the RedHat maven repository `https://maven.repository.redhat.com/ga/`. However, some applications might require some dependencies from other remote Maven repositories. In order to allow us to add another remote Maven repository, you can use `-Dts.global.s2i.maven.remote.repository=http://host:port/repo/name`. If you only want to configure different maven repositories by service, you can do it by replacing `global` to the service name, for example: `-Dts.pingPong.s2i.maven.remote.repository=...`.
```
mjurc, line 699: Small typo - Red Hat

mjurc (changes requested):
Besides that one tiny typo, this looks good

README.md
Sgitario: Fixed

Issue #85, opened 2021-06-03
S2I: ARTIFACT_COPY_ARGS is wrong for Quarkus 1.13+


mjurc:
Update the property at https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-openshift/src/main/resources/quarkus-s2i-source-build-template.yml#L35 to match the path of `target/quarkus-app/quarkus-run.jar` in the builder container.

Sgitario:
The default version will not modify the ARTIFACT_COPY_ARGS, so users would need to provide a custom template to amend this configuration. See examples/external-applications and the test OpenShiftS2iQuickstartForVersion1xIT.

PR #112, opened 2021-06-16
Allow to overwrite default template for OpenShift/Kubernetes deployments

Sgitario:
The default template used by strategies can now be overwritten using the property `ts.<GLOBAL or service name>.<kubernetes or openshift>.template`.

Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/68
Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/85

### DIFF
--- a/README.md
+++ b/README.md
@@ -666,6 +666,8 @@ public class OpenShiftPingPongResourceIT {
 }
 ```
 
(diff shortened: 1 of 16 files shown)


mjurc (approved):
Thanks for the fixes

PR #134, opened 2021-07-08
htmlunit was removed from Quarkus upstream bom

pjgg:
- Library `htmlunit` was removed from upstream bom. So we have added to our main pom.
- Quarkus command-line client doesn't support `--artifact-id` parameter in Upstream. Instead `groupId:artifactId:version` should be used. 
Example:
Quarkus 2.0.0.Final: `quarkus create app --artifact-id=app`
Quarkus Upstream: `quarkus create app example.org:app:0.0.1`

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -86,7 +86,7 @@ jobs:
           chmod +x ./quarkus-dev-cli
       - name: Build
         run: |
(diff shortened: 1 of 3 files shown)


pjgg:
htmlunit Ref: https://github.com/quarkusio/quarkus/pull/17772

PR #147, opened 2021-07-21
CLI: Use io.quarkus as platform group id

Sgitario:
- Disable flaky test in DevModeGreetingResourceIT.shouldDetectNewTests
- Disable amqp scenario because of https://github.com/quarkusio/quarkus/issues/18956
- Revert bump of HTMLunit dependency
- Bump Quarkus version to 2.0.3.Final
- CLI: Use io.quarkus as platform group id
- Remove unnecessary code from TracingExtensionExample example


.github/dependabot.yml
```diff
@@ -4,4 +4,7 @@ updates:
   - package-ecosystem: "maven"
     directory: "/"
     schedule:
-      interval: "daily"
\ No newline at end of file
+      interval: "daily"
+    ignore:
+      # Later versions of HTML Unit does not work properly when finding elements remotely.
```
rsvoboda, line 9: Any GH issue to be mentioned here ?

rsvoboda:
`Use io.quarkus as platform group id` ... what is triggering this adjustment?

Sgitario:
> `Use io.quarkus as platform group id` ... what is triggering this adjustment?

This change: https://github.com/quarkusio/quarkus/pull/18813
Before, it was using `io.quarkus` always. Now, it's using `io.quarkus.platform` that does not exist in 2.0.2.Final. 

With these changes, we ensure that we're using `io.quarkus` and the correct version we want to verify.

.github/dependabot.yml
Sgitario: Nope, this is something I experienced when running the continuous testing coverage. Nothing was working with the latest version.

Sgitario:
For the native run, we hit the same issues of some dependencies are not found. Running native with OpenShift from local works fine, So I need to investigate it a bit further in a separated PR. 
For this PR, I think we're ok for review and merge.

Sgitario:
rerun tests

Sgitario:
The error in JVM was caused by socket timeout exception (network issue). In previous runs, it worked fine.

Sgitario:
run tests

rsvoboda (approved):

Issue #148, opened 2021-07-21
Loading resources from classpath is not possible in custom build


llowinge:
Take for example amq-tcp example which uses custom build. Place eg. `myfolder/hello.txt` file under https://github.com/quarkus-qe/quarkus-test-framework/tree/main/examples/amq-tcp/src/main/resources and try it load from the app via `String s = new String(GreetingResource.class.getResourceAsStream("/myfolder/hello.txt").readAllBytes());` and it will not work. Using the same in Greetings example, it works as it uses maven build.

PR #149, opened 2021-07-21
Copy resources for custom build deployments

Sgitario:
Custom build deployments use a temporary directory at target/<service name>, so we need to copy all the resources in src/main/resources and src/test/resources, but the .properties files to have these files available at runtime.

This works for OpenShift and Kubernetes deployments as well, but not on Native deployments that should be as in https://quarkus.io/guides/writing-native-applications-tips#including-resources.

Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/148


Sgitario:
Depends on https://github.com/quarkus-qe/quarkus-test-framework/pull/147

Sgitario:
@pjgg this is ready to be reviewed.

pjgg (approved):

PR #168, opened 2021-07-30
Add key parameter to the ConsulService configuration

kshpak:
Consul key shouldn't be hardcoded.  
https://quarkus.io/guides/consul-config#configuring-the-application

This will break https://github.com/quarkus-qe/quarkus-test-suite/pull/127


Sgitario (approved):
lgtm! let's wait for the CI to be green.

PR #171, opened 2021-08-02
+Option to run containers in privileged mode

fedinskiy:
The title says it all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,7 +399,8 @@ public class GreetingResourceIT {
 }
 ```
 
-Some containers require `--privileged` mode to run properly. This mode can be enabled on a per-container basis via property `ts.{container}.container.privileged-mode=true` or for all containers via property `ts.global.container.privileged-mode=true`. This property only affects containers which are both: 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -399,6 +399,10 @@ public class GreetingResourceIT {
 }
 ```
 
+Some containers require `--privileged` mode to run properly. This mode can be enabled on a per-container basis via property `ts.{container}.container.privileged-mode=true` or for all containers via property `ts.global.container.privileged-mode=true`. This property only affects containers which are both:
```
Sgitario, line 402:
```suggestion
#### Privileged Mode
Some containers require `--privileged` mode to run properly. This mode can be enabled on a per-container basis via property `ts.<YOUR SERVICE NAME>.container.privileged-mode=true` or for all containers via property `ts.global.container.privileged-mode=true`. This property only affects containers which are both: 
```

Sgitario (approved):
lgtm! Thanks for adding this feature.
Just one minor suggestion, otherwise all good.

PR #188, opened 2021-08-06
Enable run your test without ephemeral namespaces

pjgg:
- Disable ephemeral namespaces

Most of the time ephemeral namespace is a good idea, but some other times is not possible because maybe you are not allowed to create namespaces on the fly with your own user. In those cases, you can disable ephemeral namespaces 
and run all your tests in your current namespace.

`-Dts.openshift.ephemeral.namespaces.enabled=false`

Full example command: `mvn clean verify -Popenshift -Dts.openshift.ephemeral.namespaces.enabled=false`

Pending items:
- Knative

### DIFF
--- a/README.md
+++ b/README.md
@@ -1235,6 +1235,23 @@ mvn -T 1C clean verify
 
 This Maven command would use 1 thread by CPU.
 
+- Disable ephemeral namespaces
+
+Most of the time ephemeral namespace is a good idea, but some other times is not possible because maybe you are 
+not allowed to create namespaces on the fly with your own user. In those cases, you can disable ephemeral namespaces 
+and run all your tests in your current namespace.
+
+* OpenShift: 
+`-Dts.openshift.ephemeral.namespaces.enabled=false`
+
+* Kubernetes:
+`-Dts.kubernetes.ephemeral.namespaces.enabled=false`
+
+Full example command: `mvn clean verify -Popenshift -Dts.openshift.ephemeral.namespaces.enabled=false`
+
+Default value: `ts.openshift.ephemeral.namespaces.enabled=true`
+**Note:** this feature doesn't support operators
+
 - Partial SSL support
 
 This is only supported when running tests on bare metal:
--- a/quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
+++ b/quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
@@ -11,8 +11,8 @@ import io.quarkus.test.scenarios.KubernetesScenario;
 import io.quarkus.test.utils.FileUtils;
 
 public class KubernetesExtensionBootstrap implements ExtensionBootstrap {
-
     public static final String CLIENT = "kubectl-client";
+
     private static final PropertyLookup DELETE_NAMESPACE_AFTER = new PropertyLookup("ts.kubernetes.delete.namespace.after.all",
             Boolean.TRUE.toString());
 
@@ -25,7 +25,7 @@ public class KubernetesExtensionBootstrap implements ExtensionBootstrap {
 
     @Override
     public void beforeAll(ScenarioContext context) {
-        client = KubectlClient.create();
+        client = KubectlClient.create(context.getId());
     }
 
     @Override
--- a/quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
+++ b/quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
@@ -35,17 +35,22 @@ import io.fabric8.openshift.client.NamespacedOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftConfig;
 import io.fabric8.openshift.client.OpenShiftConfigBuilder;
 import io.quarkus.test.bootstrap.Service;
+import io.quarkus.test.configuration.PropertyLookup;
 import io.quarkus.test.logging.Log;
 import io.quarkus.test.utils.Command;
 import io.quarkus.test.utils.FileUtils;
 
 public final class KubectlClient {
 
+    private static final PropertyLookup ENABLED_EPHEMERAL_NAMESPACES = new PropertyLookup(
+            "ts.kubernetes.ephemeral.namespaces.enabled", Boolean.TRUE.toString());
+
     private static final String RESOURCE_PREFIX = "resource::";
     private static final String RESOURCE_MNT_FOLDER = "/resource";
     private static final int NAMESPACE_NAME_SIZE = 10;
     private static final int NAMESPACE_CREATION_RETRIES = 5;
     private static final String LABEL_TO_WATCH_FOR_LOGS = "tsLogWatch";
+    private static final String LABEL_SCENARIO_ID = "scenarioId";
 
     private static final String KUBECTL = "kubectl";
     private static final int HTTP_PORT_DEFAULT = 80;
@@ -53,18 +58,19 @@ public final class KubectlClient {
     private final String currentNamespace;
     private final DefaultOpenShiftClient masterClient;
     private final NamespacedOpenShiftClient client;
+    private final String scenarioId;
 
-    private KubectlClient() {
-        currentNamespace = createNamespace();
-
+    private KubectlClient(String scenarioUniqueName) {
+        this.scenarioId = scenarioUniqueName;
+        String activeNamespace = new DefaultOpenShiftClient().getNamespace();
+        currentNamespace = ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean() ? createNamespace() : activeNamespace;
         OpenShiftConfig config = new OpenShiftConfigBuilder().withTrustCerts(true).withNamespace(currentNamespace).build();
-
         masterClient = new DefaultOpenShiftClient(config);
         client = masterClient.inNamespace(currentNamespace);
     }
 
-    public static KubectlClient create() {
-        return new KubectlClient();
+    public static KubectlClient create(String scenarioName) {
+        return new KubectlClient(scenarioName);
     }
 
     /**
@@ -228,13 +234,39 @@ public final class KubectlClient {
      * Delete the namespace and all the resources.
      */
     public void deleteNamespace() {
+        if (ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean()) {
+            try {
+                new Command(KUBECTL, "delete", "namespace", currentNamespace).runAndWait();
+            } catch (Exception e) {
+                fail("Project failed to be deleted. Caused by " + e.getMessage());
+            } finally {
+                masterClient.close();
+            }
+        } else {
+            deleteResourcesByLabel(LABEL_SCENARIO_ID, getScenarioId());
+        }
+    }
+
+    private String getScenarioId() {
+        return scenarioId;
+    }
+
+    /**
+     * Delete test resources.
+     */
+    private void deleteResourcesByLabel(String labelName, String labelValue) {
         try {
-            new Command(KUBECTL, "delete", "namespace", currentNamespace).runAndWait();
+            String label = String.format("%s=%s", labelName, labelValue);
+            new Command(KUBECTL, "delete", "-n", currentNamespace, "all", "-l", label).runAndWait();
         } catch (Exception e) {
             fail("Project failed to be deleted. Caused by " + e.getMessage());
         } finally {
             masterClient.close();
         }
+    }
+
+    private String extractNamespace(String namespace) {
+        return namespace.split(":")[1];
     }
 
     private boolean isPodRunning(Pod pod) {
@@ -250,6 +282,11 @@ public final class KubectlClient {
         for (HasMetadata obj : objs) {
             // set namespace
             obj.getMetadata().setNamespace(namespace());
+            Map<String, String> objMetadataLabels = Optional.ofNullable(obj.getMetadata().getLabels())
+                    .orElse(new HashMap<>());
+
+            objMetadataLabels.put(LABEL_SCENARIO_ID, getScenarioId());
+            obj.getMetadata().setLabels(objMetadataLabels);
 
             if (obj instanceof Deployment) {
                 Deployment d = (Deployment) obj;
@@ -261,7 +298,9 @@ public final class KubectlClient {
                 d.getSpec().getTemplate().getMetadata().setNamespace(namespace());
 
                 // add label for logs
-                d.getSpec().getTemplate().getMetadata().getLabels().put(LABEL_TO_WATCH_FOR_LOGS, service.getName());
+                Map<String, String> templateMetadataLabels = d.getSpec().getTemplate().getMetadata().getLabels();
+                templateMetadataLabels.put(LABEL_TO_WATCH_FOR_LOGS, service.getName());
+                templateMetadataLabels.put(LABEL_SCENARIO_ID, getScenarioId());
 
                 // add env var properties
                 Map<String, String> enrichProperties = enrichProperties(service.getProperties(), d);
--- a/quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
+++ b/quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
@@ -55,6 +55,7 @@ import io.fabric8.openshift.client.NamespacedOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftConfig;
 import io.fabric8.openshift.client.OpenShiftConfigBuilder;
 import io.quarkus.test.bootstrap.Service;
+import io.quarkus.test.configuration.PropertyLookup;
 import io.quarkus.test.logging.Log;
 import io.quarkus.test.services.operator.model.CustomResourceStatus;
 import io.quarkus.test.utils.Command;
@@ -64,6 +65,7 @@ import okhttp3.Response;
 public final class OpenShiftClient {
 
     public static final String LABEL_TO_WATCH_FOR_LOGS = "tsLogWatch";
+    public static final String LABEL_SCENARIO_ID = "scenarioId";
 
     private static final String IMAGE_STREAM_TIMEOUT = "imagestream.install.timeout";
     private static final String OPERATOR_INSTALL_TIMEOUT = "operator.install.timeout";
@@ -79,23 +81,27 @@ public final class OpenShiftClient {
 
     private static final String OC = "oc";
 
+    private static final PropertyLookup ENABLED_EPHEMERAL_NAMESPACES = new PropertyLookup(
+            "ts.openshift.ephemeral.namespaces.enabled", Boolean.TRUE.toString());
+
     private final String currentNamespace;
     private final DefaultOpenShiftClient masterClient;
     private final NamespacedOpenShiftClient client;
     private final KnativeClient kn;
+    private final String scenarioId;
 
-    private OpenShiftClient() {
-        currentNamespace = createProject();
-
+    private OpenShiftClient(String scenarioId) {
+        this.scenarioId = scenarioId;
+        String activeNamespace = new DefaultOpenShiftClient().getNamespace();
+        currentNamespace = ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean() ? createProject() : activeNamespace;
         OpenShiftConfig config = new OpenShiftConfigBuilder().withTrustCerts(true).withNamespace(currentNamespace).build();
-
         masterClient = new DefaultOpenShiftClient(config);
         client = masterClient.inNamespace(currentNamespace);
         kn = client.adapt(KnativeClient.class);
     }
 
-    public static OpenShiftClient create() {
-        return new OpenShiftClient();
+    public static OpenShiftClient create(String scenarioId) {
+        return new OpenShiftClient(scenarioId);
     }
 
     /**
@@ -217,8 +223,8 @@ public final class OpenShiftClient {
         }
 
         try {
-            new Command(OC, "expose", "svc/" + serviceName, "--port=" + port, "-n", currentNamespace)
-                    .runAndWait();
+            new Command(OC, "expose", "svc/" + serviceName, "--port=" + port, "-n", currentNamespace,
+                    "-l " + LABEL_SCENARIO_ID + "=" + getScenarioId()).runAndWait();
         } catch (Exception e) {
             fail("Service failed to be exposed. Caused by " + e.getMessage());
         }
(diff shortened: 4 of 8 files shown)


quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -41,30 +44,40 @@
...
 public final class KubectlClient {
 
+    public static final String K8S_LABEL_SCENARIO_ID = "scenarioId";
+
     private static final String RESOURCE_PREFIX = "resource::";
     private static final String RESOURCE_MNT_FOLDER = "/resource";
     private static final int NAMESPACE_NAME_SIZE = 10;
     private static final int NAMESPACE_CREATION_RETRIES = 5;
-    private static final String LABEL_TO_WATCH_FOR_LOGS = "tsLogWatch";
+    private static final String K8S_LABEL_TO_WATCH_FOR_LOGS = "tsLogWatch";
```
Sgitario, line 53: Why adding a `K8S_` prefix in this field? I think taking into account that this is in the KubectlClient, we don't need it.
Also, group the label fields together.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -41,30 +44,40 @@
...
     private final String currentNamespace;
     private final DefaultOpenShiftClient masterClient;
     private final NamespacedOpenShiftClient client;
-
-    private KubectlClient() {
-        currentNamespace = createNamespace();
+    private final String scenarioId;
+    private final Pattern namespaceRegexp = Pattern.compile(".*namespace.*");
+
+    private KubectlClient(boolean ephemeralNamespaceEnabled) {
```
Sgitario, line 65: Having boolean parameters to do one or another thing is generally a bad practice. Moreover, maybe this logic should be now in KubernetesExtensionBootstrap.java.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -230,6 +243,48 @@ public void deleteNamespace() {
         }
     }
 
+    /**
+     * Delete test resources.
+     */
+    public void deleteResourcesByLabel(String labelName, String labelValue) {
```
Sgitario, line 249: I think this method should not be opened in the KubectlClient API (this API is open to be used for users).

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -377,4 +435,7 @@ private void printServiceInfo(Service service) {
         }
     }
 
+    private String generateScenarioId() {
+        return "" + System.currentTimeMillis();
```
Sgitario, line 439: This might return a different ID for the same test resources... 
Please, use the test class name instead.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/services/containers/KubernetesContainerManagedResource.java
```diff
@@ -104,7 +104,8 @@ private String replaceDeploymentContent(String content) {
 
         return content.replaceAll(quote("${IMAGE}"), model.getImage())
                 .replaceAll(quote("${SERVICE_NAME}"), model.getContext().getName())
-                .replaceAll(quote("${INTERNAL_PORT}"), "" + model.getPort());
+                .replaceAll(quote("${INTERNAL_PORT}"), "" + model.getPort())
+                .replaceAll(quote("${TEST_CASE_NAME}"), "" + client.getScenarioId());
```
Sgitario, line 108: Why do we need to do this? You're already adding the label in `enrichTemplate`.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/services/quarkus/KubernetesQuarkusApplicationManagedResource.java
```diff
@@ -130,7 +130,8 @@ private String replaceDeploymentContent(String content) {
                 .replaceAll(quote("${SERVICE_NAME}"), model.getContext().getName())
                 .replaceAll(quote("${ARTIFACT}"), model.getArtifact().getFileName().toString())
                 .replaceAll(quote("${INTERNAL_PORT}"),
-                        model.getContext().getOwner().getProperty(QUARKUS_HTTP_PORT_PROPERTY, "" + HTTP_PORT_DEFAULT));
+                        model.getContext().getOwner().getProperty(QUARKUS_HTTP_PORT_PROPERTY, "" + HTTP_PORT_DEFAULT))
+                .replaceAll(quote("${TEST_CASE_NAME}"), client.getScenarioId());
```
Sgitario, line 134: Why do we need to do this? You're already adding the label in `enrichTemplate`.

quarkus-test-kubernetes/src/main/resources/kubernetes-deployment-template.yml
```diff
@@ -6,6 +6,8 @@ items:
   kind: "Deployment"
   metadata:
     name: "${SERVICE_NAME}"
+    labels:
+      scenarioId: "${TEST_CASE_NAME}"
```
Sgitario, line 10: Why do we need to do this? You're already adding the label in `enrichTemplate`.

quarkus-test-kubernetes/src/main/resources/kubernetes-deployment-template.yml
```diff
@@ -15,6 +17,7 @@ items:
       metadata:
         labels:
           deployment: "${SERVICE_NAME}"
+          scenarioId: "${TEST_CASE_NAME}"
```
Sgitario, line 20: Why do we need to do this? You're already adding the label in `enrichTemplate`.

README.md
```diff
@@ -1141,6 +1141,18 @@ mvn -T 1C clean verify
 
 This Maven command would use 1 thread by CPU.
 
+- Disable ephemeral namespaces
+
+Most of the time ephemeral namespace is a good idea, but some other times is not possible because maybe you are 
+not allowed to create namespaces on the fly with your own user. In those cases, you can disable ephemeral namespaces 
+and run all your tests in your current namespace.
+
+`-Dts.openshift.ephemeral.namespaces.enabled=false`
```
Sgitario, line 1150: Add a note about the K8s property too.

Sgitario (changes requested):
The same comments from Kubernetes' changes apply also to OpenShift's changes.
In general:
- The label ID is using a timestamp that might not be unique. Use the test class name instead.
- We don't need to update the templates but add the label in the `enrichTemplate` method only (as it has already been done).
- The current solution does not work when adding custom resources in Operators (as it would require users to manually add the label).
- Move the logic of create/delete resources from the K8s/OCP API clients to the bootstrap, so we prevent users from using this logic.

Moreover, I'm a bit reluctant to:
- Support this feature as it embraces a testing bad practice. An environment must be repeatable and reproducible and we can't guarantee that all the resources will be cleared out (see custom resources or resources that users will create as part of the test scenario). Still, I see benefit though.
- Keep adding new features on the test framework when we didn't run a fully green execution of our test suite (even for JVM and Native). Even more, we don't have an OCP environment stable to verify the test framework yet. 

I would like to open the discussion to @rsvoboda and @mjurc too.

README.md
```diff
@@ -1141,6 +1141,22 @@ mvn -T 1C clean verify
 
 This Maven command would use 1 thread by CPU.
 
+- Disable ephemeral namespaces
+
+Most of the time ephemeral namespace is a good idea, but some other times is not possible because maybe you are 
+not allowed to create namespaces on the fly with your own user. In those cases, you can disable ephemeral namespaces 
+and run all your tests in your current namespace.
+
+* Ocp:
```
Sgitario, line 1150:
```suggestion
* OpenShift: 
```

README.md
```diff
@@ -1141,6 +1141,22 @@ mvn -T 1C clean verify
...
+- Disable ephemeral namespaces
+
+Most of the time ephemeral namespace is a good idea, but some other times is not possible because maybe you are 
+not allowed to create namespaces on the fly with your own user. In those cases, you can disable ephemeral namespaces 
+and run all your tests in your current namespace.
+
+* Ocp: 
+`-Dts.openshift.ephemeral.namespaces.enabled=false`
+
+* K8s:
```
Sgitario, line 1153:
```suggestion
* Kubernetes:
```

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -243,6 +296,7 @@ private String enrichTemplate(Service service, String template) {
         for (HasMetadata obj : objs) {
             // set namespace
             obj.getMetadata().setNamespace(namespace());
+            obj.getMetadata().setLabels(Map.of(LABEL_SCENARIO_ID, getScenarioId()));
```
Sgitario, line 299: If the metadata had labels already, using `setLabels` would overwrite it.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -525,7 +562,7 @@ private String enrichTemplate(Service service, String template) {
         for (HasMetadata obj : objs) {
             // set namespace
             obj.getMetadata().setNamespace(project());
-
+            obj.getMetadata().setLabels(Map.of(LABEL_SCENARIO_ID, getScenarioId()));
```
Sgitario, line 565: If the metadata had labels already, using `setLabels` would overwrite it.

quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/ExtensionOpenShiftQuarkusApplicationManagedResource.java
```diff
@@ -116,6 +117,11 @@ private String withLabelsForWatching() {
                 model.getContext().getOwner().getName());
     }
 
+    //    private String withLabelsForscenarioId() {
+    //        return withProperty(QUARKUS_OPENSHIFT_LABELS + OpenShiftClient.LABEL_SCENARIO_ID,
+    //                client.getScenarioId());
+    //    }
+
```
Sgitario, line 124: ???

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -230,6 +241,48 @@ public void deleteNamespace() {
         }
     }
 
+    private String findActiveNamespace() {
+        List<String> namespaceOutputResp = new ArrayList<>();
```
Sgitario, line 245: There is a much easier way to get the current namespace: `return new DefaultOpenShiftClient().getNamespace();`

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -35,36 +38,43 @@
...
     private final NamespacedOpenShiftClient client;
+    private final String scenarioId;
+    private final Pattern namespaceRegexp = Pattern.compile(".*namespace.*");
 
-    private KubectlClient() {
-        currentNamespace = createNamespace();
-
+    private KubectlClient(String scenarioUniqueName) {
+        this.scenarioId = scenarioUniqueName;
+        currentNamespace = (ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean()) ? createNamespace() : findActiveNamespace();
```
Sgitario, line 70: There is extra paranthesis here.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -79,23 +81,26 @@
...
     private final NamespacedOpenShiftClient client;
     private final KnativeClient kn;
+    private final String scenarioId;
 
-    private OpenShiftClient() {
-        currentNamespace = createProject();
-
+    private OpenShiftClient(String scenarioUniqueName) {
+        this.scenarioId = scenarioUniqueName;
+        currentNamespace = (ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean()) ? createProject() : findActiveNamespace();
```
Sgitario, line 95: There is extra paranthesis here.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -511,8 +516,40 @@ public String getStatus() {
      * Delete the current project and all its resources.
      */
     public void deleteProject() {
+        deleteResourcesByLabel(LABEL_SCENARIO_ID, getScenarioId());
```
Sgitario, line 519: If ephemeral is active, this is not necessary. Let's remove the resources by label only if ephemeral is disabled.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -511,8 +516,40 @@ public String getStatus() {
...
+            } catch (Exception e) {
+                fail("Project failed to be deleted. Caused by " + e.getMessage());
+            } finally {
+                masterClient.close();
+            }
+        }
+    }
+
+    private String findActiveNamespace() {
+        List<String> namespaceOutputResp = new ArrayList<>();
```
Sgitario, line 532: The command `oc project -q` is not working for me when using OCP 4.8 (not sure if it's an intermittent issue or not). But there is a much easier way to get the active namespace: `return new DefaultOpenShiftClient().getNamespace();` and as far I see, it's working consistently.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -511,8 +516,40 @@ public String getStatus() {
...
-            new Command(OC, "delete", "project", client.getNamespace()).runAndWait();
+            new Command(OC, "project", "-q").outputToLines(namespaceOutputResp).runAndWait();
+        } catch (Exception e) {
+            fail("Retrieve namespace has failed. Caused by " + e.getMessage());
+        }
+
+        return namespaceOutputResp.stream().findFirst().orElseThrow(RuntimeException::new);
+    }
+
+    public String getScenarioId() {
```
Sgitario, line 542: this should be private

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -221,6 +231,7 @@ public int port(Service service) {
      * Delete the namespace and all the resources.
      */
     public void deleteNamespace() {
+        deleteResourcesByLabel(LABEL_SCENARIO_ID, getScenarioId());
```
Sgitario, line 234: We should only delete resources by label when the ephemeral namespaces flag is disabled.

Sgitario (changes requested):
- Labels are being overwritten by the scenario ID label
- To get the active namespace: the command `oc project -q` is not working for me when using OCP 4.8 (not sure if it's an intermittent issue or not). But there is a much easier way to get the active namespace: `return new DefaultOpenShiftClient().getNamespace();` and as far I see, it's working consistently.
- The OpenShift extension deployment strategy is not adding the scenario label and hence is not being removed.
- The operators are not being deleted either.

quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/ExtensionOpenShiftQuarkusApplicationManagedResource.java
```diff
@@ -101,6 +101,7 @@ private void deployProjectUsingMavenCommand() {
         args.add(withKubernetesClientTrustCerts());
         args.add(withContainerImageGroup(namespace));
         args.add(withLabelsForWatching());
+        //   args.add(withLabelsForscenarioId());
```
Sgitario, line 104: We should add the labels to the Quarkus OpenShift extension, otherwise it won't work for this deployment strategy.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -511,8 +517,30 @@ public String getStatus() {
...
+    }
+
+    /**
+     * Delete test resources.
+     */
+    private void deleteResourcesByLabel(String labelName, String labelValue) {
         try {
-            new Command(OC, "delete", "project", client.getNamespace()).runAndWait();
+            String label = String.format("%s=%s", labelName, labelValue);
+            new Command(OC, "delete", "-n", client.getNamespace(), "all", "-l", label).runAndWait();
```
Sgitario, line 543: I think it's safer to use `currentNamespace` instead of `client.getNamespace()`

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -511,8 +517,30 @@ public String getStatus() {
      * Delete the current project and all its resources.
      */
     public void deleteProject() {
+        if (ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean()) {
+            try {
+                new Command(OC, "delete", "project", client.getNamespace()).runAndWait();
```
Sgitario, line 522: I think it's safer to use `currentNamespace` instead of `client.getNamespace()`

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -221,15 +229,41 @@ public int port(Service service) {
...
+    }
+
+    /**
+     * Delete test resources.
+     */
+    private void deleteResourcesByLabel(String labelName, String labelValue) {
         try {
-            new Command(KUBECTL, "delete", "namespace", currentNamespace).runAndWait();
+            String label = String.format("%s=%s", labelName, labelValue);
+            new Command(KUBECTL, "delete", "-n", client.getNamespace(), "all", "-l", label).runAndWait();
```
Sgitario, line 255: I think it's safer to use `currentNamespace` instead of `client.getNamespace()`

Sgitario (changes requested):
The only missing important bit are:
- The OpenShift extension deployment strategy is not adding the scenario label and hence is not being removed.
- Operators are not being uninstalled
- Custom Resources (for operators) are not being uninstalled either

For the operators example, please try with `OpenShiftStrimziOperatorKafkaWithoutRegistryMessagingIT` scenario for example to double check everything is cleared out.

pjgg:
This feature doesn't support Operators.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -35,36 +35,44 @@
...
     private static final String RESOURCE_PREFIX = "resource::";
     private static final String RESOURCE_MNT_FOLDER = "/resource";
     private static final int NAMESPACE_NAME_SIZE = 10;
     private static final int NAMESPACE_CREATION_RETRIES = 5;
     private static final String LABEL_TO_WATCH_FOR_LOGS = "tsLogWatch";
+    private static final String LABEL_SCENARIO_ID = "scenarioId";
 
     private static final String KUBECTL = "kubectl";
     private static final int HTTP_PORT_DEFAULT = 80;
+    private static final int KUBECTL_CMD_TIMEOUT_SEC = 5;
```
Sgitario, line 57: This field is not used.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -35,36 +35,44 @@
...
 
     private final String currentNamespace;
     private final DefaultOpenShiftClient masterClient;
     private final NamespacedOpenShiftClient client;
+    private final String scenarioId;
+    private final Pattern namespaceRegexp = Pattern.compile(".*namespace.*");
```
Sgitario, line 63: This field is not used.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
```diff
@@ -536,7 +573,9 @@ private String enrichTemplate(Service service, String template) {
                 dc.getSpec().getTemplate().getMetadata().setNamespace(project());
 
                 // add label for logs
```
Sgitario, line 575: Update comment to reflect the new behaviour

Sgitario (changes requested):
Some leftover and one minor change (about the comment).
Once updated, I think we can proceed with the test on OCP.

Sgitario:
run tests

Sgitario:
run tests - with ephemeral namespaces disabled.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
```diff
@@ -27,7 +28,7 @@ public boolean appliesFor(ExtensionContext context) {
 
     @Override
     public void beforeAll(ExtensionContext context) {
-        client = KubectlClient.create();
+        client = KubectlClient.create(generateScenarioId(context));
```
Sgitario, line 31: To be refactored in https://github.com/quarkus-qe/quarkus-test-framework/issues/209

Sgitario (approved):
Let's wait for the CI before merging

Sgitario:
run tests

Sgitario:
run tests

Sgitario:
run tests

Sgitario (changes requested):
Failed the OCP jobs due that some services generated by serverless are not cleaned up and hence doing the following scenarios to fail because the route host is already claimed:

```
oc: route/app not accepted: HostAlreadyClaimed (svc/app)
```

Sgitario (changes requested):
Remove leftovers in the OCP and K8S extension bootstraps.

Sgitario (approved):

PR #193, opened 2021-08-10
[WIP] Include MacOS environment in Github CI

Sgitario:
TODO: If PR CI works, add coverage of macos in JVM and Native as part of the Daily run.


Sgitario:
@rsvoboda I will close this PR as I don't think we need MacOS coverage (macos-11 is not working fine).

rsvoboda:
OK

PR #203, opened 2021-08-12
Fast fail a Quarkus application that fails with "Failed to load config value of type class"

Sgitario:


rsvoboda:
Was `.github/workflows/kubernetes.yaml` included intentionally ?

Sgitario:
> Was `.github/workflows/kubernetes.yaml` included intentionally ?

I used it to check why Kubernetes deployment were failing. It's deleted now.

PR #207, opened 2021-08-17
Add reactive webclient service

Sgitario:
We will use the Vert.x Munity webclient implementation
These changes are being covered in the Greetings example and in OpenShift and K8s.

### DIFF
--- a/examples/greetings/src/test/java/io/quarkus/qe/ReactiveGreetingResourceIT.java
+++ b/examples/greetings/src/test/java/io/quarkus/qe/ReactiveGreetingResourceIT.java
@@ -1,0 +1,31 @@
+package io.quarkus.qe;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.net.HttpURLConnection;
+
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.QuarkusApplication;
+import io.vertx.mutiny.ext.web.client.HttpResponse;
+import io.vertx.mutiny.ext.web.client.predicate.ResponsePredicate;
+
+@QuarkusScenario
+public class ReactiveGreetingResourceIT {
+
+    @QuarkusApplication
+    static final RestService app = new RestService();
+
+    @Test
+    public void shouldSayDefaultGreeting() {
+        String response = app.mutiny().get("/reactive-greeting")
+                .expect(ResponsePredicate.status(HttpURLConnection.HTTP_OK))
+                .send()
+                .map(HttpResponse::bodyAsString)
+                .await().indefinitely();
+
+        assertEquals("Hello, I'm victor", response);
+    }
+}
--- a/quarkus-test-core/pom.xml
+++ b/quarkus-test-core/pom.xml
@@ -53,5 +53,9 @@
             <groupId>io.prometheus</groupId>
             <artifactId>simpleclient_pushgateway</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye.reactive</groupId>
+            <artifactId>smallrye-mutiny-vertx-web-client</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/RestService.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/RestService.java
@@ -1,11 +1,18 @@
 package io.quarkus.test.bootstrap;
 
+import org.apache.commons.lang3.StringUtils;
+
 import io.restassured.RestAssured;
 import io.restassured.specification.RequestSpecification;
+import io.vertx.ext.web.client.WebClientOptions;
+import io.vertx.mutiny.core.Vertx;
+import io.vertx.mutiny.ext.web.client.WebClient;
 
 public class RestService extends BaseService<RestService> {
 
     private static final String BASE_PATH = "/";
+
+    private WebClient webClient;
 
     public RequestSpecification given() {
         return RestAssured.given().baseUri(getHost()).basePath(BASE_PATH).port(getPort());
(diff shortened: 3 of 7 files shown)


quarkus-test-core/pom.xml
```diff
@@ -53,5 +53,9 @@
             <groupId>io.prometheus</groupId>
             <artifactId>simpleclient_pushgateway</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye.reactive</groupId>
+            <artifactId>smallrye-mutiny-vertx-web-client</artifactId>
```
Sgitario, line 58: @pjgg do you foresee any side effects of adding this dependency?

pjgg: No problem at all

examples/greetings/src/main/java/io/quarkus/qe/GreetingResource.java
```diff
@@ -9,6 +9,9 @@
 
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
+import io.smallrye.common.annotation.Blocking;
+
+@Blocking
```
pjgg, line 14: `@Blocking` is not really necessary if you return `T` or `void` .

examples/greetings/src/test/java/io/quarkus/qe/ReactiveGreetingResourceIT.java
```diff
@@ -0,0 +1,31 @@
...
+    @Test
+    public void shouldSayDefaultValue() {
+        String response = app.webClient().get("/reactive-greeting")
+                .expect(ResponsePredicate.status(HttpURLConnection.HTTP_OK))
+                .send()
+                .map(HttpResponse::bodyAsString)
+                .await().indefinitely();
+
+        assertEquals("Hello, I'm victor", response);
+    }
```
pjgg, line 30: instead of swap from non-blocking to blocking we could also test it in a reactive way:

````
@QuarkusScenario
public class ReactiveGreetingResourceIT {

    @QuarkusApplication
    static final ReactiveWebClientService app = new ReactiveWebClientService();

    @Test
    public void shouldSayDefaultValue() {

        UniAssertSubscriber<String> subscriber = app.webClient().get("/reactive-greeting")
                .expect(ResponsePredicate.status(HttpURLConnection.HTTP_OK))
                .send()
                .map(HttpResponse::bodyAsString)
                .subscribe().withSubscriber(UniAssertSubscriber.create());

        subscriber.awaitItem(Duration.ofSeconds(10))
                .assertCompleted()
                .assertItem("Hello, I'm victor");
    }
}
````

File: quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/ReactiveWebClientService.java [hunk 4]
```diff
@@ -0,0 +1,31 @@
+package io.quarkus.test.bootstrap;
+
+import org.apache.commons.lang3.StringUtils;
+
+import io.vertx.ext.web.client.WebClientOptions;
+import io.vertx.mutiny.core.Vertx;
+import io.vertx.mutiny.ext.web.client.WebClient;
+
+public class ReactiveWebClientService extends BaseService<ReactiveWebClientService> {
```
pjgg, line 9: What about renaming the service. I mean, is reactive that is true but could be RxJava, or `callback-hell` or...Mutiny!
By this way would be more aligned to the `RestService`

In my opinion, I will rename this class to `RestMutinyService` but also I will rename `RestService` to `RestAssuranceService`. 

Why `RestMutinyService` ?, we are using under the hood vertx that is true, but vertx has a lot of flavors, so ReactiveWebClient is too open... we are talking about Vertx futures, rxJava, Mutiny... ? 

And I think that `RestService` could be renamed to `RestAssuranceService` following the same principle. 

On the other hand (going back to `ReactiveWebClientService` class) I think that we could rename the method `webClient` by `intance`

Something like this: 

```
public class RestMutinyService extends BaseService<RestMutinyService> {

    private WebClientOptions webClientOptions = new WebClientOptions();
    private WebClient webClient;

    public RestMutinyService withWebClientOptions(WebClientOptions options) {
        this.webClientOptions = options;
        return this;
    }

    @Override
    public void start() {
        super.start();

        webClient = WebClient.create(Vertx.vertx(), webClientOptions
                .setDefaultHost(getHost().replace("http://", StringUtils.EMPTY))
                .setDefaultPort(getPort()));
    }

    public WebClient instance() {
        return webClient;
    }
}

```


And then could be used by this way:

```

@QuarkusScenario
public class ReactiveGreetingResourceIT {

    @QuarkusApplication
    static final RestMutinyService app = new RestMutinyService();

    @Test
    public void shouldSayDefaultValue() {

        UniAssertSubscriber<String> subscriber = app.instance().get("/reactive-greeting")
                .expect(ResponsePredicate.status(HttpURLConnection.HTTP_OK))
                .send()
                .map(HttpResponse::bodyAsString)
                .subscribe().withSubscriber(UniAssertSubscriber.create());

        subscriber.awaitItem(Duration.ofSeconds(10))
                .assertCompleted()
                .assertItem("Hello, I'm victor");
    }
}

```

pjgg (changes requested):

examples/greetings/src/main/java/io/quarkus/qe/GreetingResource.java
Sgitario: Yes, I know, but I wanted to explicitly say that it was blocking.
Anyway, I'm going to remove it to avoid misconceptions.

examples/greetings/src/test/java/io/quarkus/qe/ReactiveGreetingResourceIT.java
Sgitario: I see the reactive way much more complicated than the blocking way, but I've changed following your suggestion.

quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/ReactiveWebClientService.java
Sgitario: I don't like to rename `webClient` to `instance`, the final usage would be a bit odd: `app.instance()...`.
However I think you made a good point about the REST service implementations, but what about easing this by removing the new service called `ReactiveWebClientService` and moving the `WebClient` stuff to the existing RestService? 
This way in RestService we have the blocking approach (RestAssured) and the reactive way (using Vert.x Mutiny client) to interact with our services.
I've made this change already, so looking forward to hearing from you.

examples/greetings/src/test/java/io/quarkus/qe/ReactiveGreetingResourceIT.java
Sgitario: The reactive way is failing for Quarkus 1.13.7.Final (the mutiny API has changed).
I will keep using the original approach.

quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/RestService.java
```diff
@@ -15,6 +22,20 @@ public RequestSpecification https() {
         return RestAssured.given().baseUri(getHost(Protocol.HTTPS)).basePath("/").port(getPort(Protocol.HTTPS));
     }
 
+    public WebClient mutiny() {
+        return mutiny(new WebClientOptions());
+    }
+
+    public WebClient mutiny(WebClientOptions options) {
```
pjgg, line 29: brilliant!, too much better by this way!

pjgg (approved):

Issue #206, opened 2021-08-17
S2i: Native support of Quarkus Application from Git repositories


Sgitario:
Having:

```
@OpenShiftScenario
public class OpenShiftQuickstartUsingS2iIT {
    @GitRepositoryQuarkusApplication(repo = "https://github.com/quarkusio/quarkus-quickstarts.git", contextDir = "getting-started")
    static final RestService app = new RestService();
```

It does not work for Native builds.

rsvoboda:
Please check what image was used in OpenShiftTS for S2I in native mode.

PR #208, opened 2021-08-18
Support S2i for Native builds

Sgitario:
In order to manage the base image in use, you need to provide the properties:
- For JVM: `ts.global.s2i.quarkus.jvm.base.image`
- For Native: `ts.global.s2i.quarkus.native.base.image`

The way these properties are up to users. In the examples, we supply this configuration in the pom.xml as part of system properties (in the Maven failsafe plugin). 
But we can provide a custom property by service in the `test.properties` file. For further information about how to customise the properties, go to the Configuration section.

Fix https://github.com/quarkus-qe/quarkus-test-framework/issues/206


README.md
```diff
@@ -784,6 +784,13 @@ public class OpenShiftS2iQuickstartIT {
     //
 ```
 
+This scenario will work for JVM and Native builds. In order to manage the base image in use, you need to provide the properties:
+- For JVM: `ts.global.s2i.quarkus.jvm.base.image`
```
mjurc, line 788: The only thing I mind in this PR is naming. Maybe the property should be called `ts.global.s2i.quarkus.jvm.builder.image` rather than `ts.global.s2i.quarkus.jvm.base.image`? Those are builder images after all.

mjurc (changes requested):
Otherwise, LGTM

README.md
Sgitario: Changed

mjurc (approved):
LGTM.

Can you squash the commits?

Sgitario:
> LGTM.
> 
> Can you squash the commits?

Ops! Squashed.

mjurc (approved):

PR #211, opened 2021-08-19
Add some DocketUtils methods in order to remove and retrieve images

pjgg:
Add the following methods to DockerUtils:

* getImage: Returns an image based on the provided image name and version
* removeImage: Returns true if docker image is removed
* isVersion: Returns true if docker images contains expected version

### DIFF
--- a/quarkus-test-images/pom.xml
+++ b/quarkus-test-images/pom.xml
@@ -13,5 +13,9 @@
             <groupId>io.quarkus.qe</groupId>
             <artifactId>quarkus-test-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.docker-java</groupId>
+            <artifactId>docker-java</artifactId>
+        </dependency>
     </dependencies>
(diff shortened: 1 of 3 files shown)


quarkus-test-images/pom.xml
```diff
@@ -13,5 +13,9 @@
             <groupId>io.quarkus.qe</groupId>
             <artifactId>quarkus-test-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
```
Sgitario, line 18: I would prefer not adding the testcontainer dependency in here (it should be only in quarkus-test-containers).
We should be able to use the https://github.com/docker-java/docker-java directly.

Sgitario (changes requested):
Drop the testcontainers dependency and use directly the docker-java API.

Sgitario (approved):

PR #215, opened 2021-08-23
Refactor Contexts: Create ScenarioContext that wraps ExtensionContext

Sgitario:
ExtensionContext comes from JUnit and we were coupling it in too many interfaces.
With this refactor, we have a scenario context that wraps the ExtensionContext and some common functionality like:
- get / check annotations present
- common test store (junit)
- generate an unique id per test scenario

Moreover, we have created a ServiceListener to bind actions to the service lifespan (start, stop, error).

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/ScenarioContext.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/ScenarioContext.java
@@ -1,0 +1,67 @@
+package io.quarkus.test.bootstrap;
+
+import java.lang.annotation.Annotation;
+import java.util.Optional;
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import io.quarkus.test.logging.Log;
+
+public final class ScenarioContext {
+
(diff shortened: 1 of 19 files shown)


quarkus-test-core/src/main/java/io/quarkus/test/bootstrap/ScenarioContext.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public <T extends Annotation> T getAnnotation(Class<T> annotationClass) {
+        return getTestContext().getRequiredTestClass().getAnnotation(annotationClass);
+    }
+
+    public void setMethodTestContext(ExtensionContext methodTestContext) {
+        this.methodTestContext = methodTestContext;
+    }
+
+    private static String generateScenarioId(ExtensionContext context) {
```
pjgg, line 61: I think that this id could follow this pattern: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

id doesn't hurt  and then later could be used in several places as Labels.

pjgg (approved):

PR #217, opened 2021-08-23
Support assertion for startup failures

Sgitario:
With the test framework, we can assert startup failures using `service.setAutoStart(false)`. When disabling this flag, the
test framework will not start the service and users will need to manually start them by doing `service.start()` at each test case. 
Hence users should be able now to assert failure messages from the logs for each test case. For example:

```java
@QuarkusApplication
static final RestService app = new RestService()
        .setAutoStart(false);

@Test
public void shouldFailOnStart() {
    assertThrows(AssertionError.class, () -> app.start(),
            "Should fail because runtime exception in ValidateCustomProperty");
    // or checks service logs
    app.logs().assertContains("Missing property a.b.z");
}
```

Moreover, we can try to fix the application during the test execution:

```java
@Test
public void shouldWorkWhenPropertyIsCorrect() {
    app.withProperty("a.b.z", "here you have!");
    app.start();
    app.given().get("/hello").then().statusCode(HttpStatus.SC_OK);
}
```

### DIFF
--- a/examples/greetings/src/main/java/io/quarkus/qe/ValidateCustomProperty.java
+++ b/examples/greetings/src/main/java/io/quarkus/qe/ValidateCustomProperty.java
@@ -1,0 +1,27 @@
+package io.quarkus.qe;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.event.Observes;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+import io.quarkus.runtime.StartupEvent;
+
+@ApplicationScoped
+public class ValidateCustomProperty {
+
+    public static final String DISALLOW_PROPERTY_VALUE = "WRONG!";
+    public static final String CUSTOM_PROPERTY = "custom.property.name";
+
+    private static final Logger LOG = Logger.getLogger(ValidateCustomProperty.class.getName());
+
+    @ConfigProperty(name = CUSTOM_PROPERTY)
+    String value;
+
+    void onStart(@Observes StartupEvent ev) {
+        if (DISALLOW_PROPERTY_VALUE.equals(value)) {
+            throw new RuntimeException("Wrong value! " + value);
+        }
+    }
+}
--- a/README.md
+++ b/README.md
@@ -133,6 +133,37 @@ class MyScenarioIT extends MyParent {
 
 Now, the framework will initialize the `appInChild` service first and then the `appInParent` service.
 
+### Test expected failures
+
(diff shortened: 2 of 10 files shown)


Sgitario:
Depends on https://github.com/quarkus-qe/quarkus-test-framework/pull/215

examples/greetings/src/main/java/io/quarkus/qe/ValidateCustomProperty.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+import io.quarkus.runtime.StartupEvent;
+
+@ApplicationScoped
+public class ValidateCustomProperty {
+
+    public static final String DISALLOW_PROPERTY = "WRONG!";
```
kshpak, line 14:
```suggestion
    public static final String DISALLOW_PROPERTY_VALUE = "WRONG!";
```

examples/greetings/src/main/java/io/quarkus/qe/ValidateCustomProperty.java
```diff
@@ -0,0 +1,28 @@
...
+    public static final String CUSTOM_PROPERTY = "custom.property.name";
+
+    private static final Logger LOG = Logger.getLogger(ValidateCustomProperty.class.getName());
+
+    @ConfigProperty(name = CUSTOM_PROPERTY)
+    String value;
+
+    void onStart(@Observes StartupEvent ev) {
+        LOG.info("DASD: " + value);
```
kshpak, line 23: what is DASD?

rsvoboda:
From the description this looks really nice. Is this "bare metal" specific feature or there is support for OpenShift / Kubernetes targets too?

Sgitario:
> From the description this looks really nice. Is this "bare metal" specific feature or there is support for OpenShift / Kubernetes targets too?

It's supported also for OpenShift and Kubernetes.

examples/greetings/src/main/java/io/quarkus/qe/ValidateCustomProperty.java
Sgitario: Updated

examples/greetings/src/main/java/io/quarkus/qe/ValidateCustomProperty.java
Sgitario: Ups!
Deleted.

kshpak (approved):

PR #221, opened 2021-08-25
Refactor solution for Metrics ++ Polish properties

Sgitario:
The changes in this PR are:
- Change the tracing/metrics format to follow the same format as the rest of the properties

This change allows to enable/disable/configure these extensions per module.

- Refactor the metrics module to decouple the metrics of prometheus and have another type of exporter: file.

My idea is to integrate the metrics within the running Quarkus application, so we can assert metrics during the test execution without needing to have a Prometheus.

Also, we're exporting the metrics into a file (/target/logs/metrics.out).


misc/Metrics.md
```diff
@@ -1,36 +1,34 @@
...
-   - `ts_quarkus_gauge_requests_fail` represent the total amount of failures 
-   - `ts_quarkus_gauge_modules` represents metrics associated to the test modules. You will have one of this metrics per quarkus scenario / module.
-        - `ts_quarkus_module_name` represents the maven modules name
-        - `ts_quarkus_module_status` could be `module_success` or `module_fail`
-   - `ts_quarkus_histogram_modules_duration`
+   - `tests_total` represent the total amount of test
+   - `tests_success` represent the total amount of succeed test
+   - `tests_ignored` represent the total amount of ignored test
+   - `tests_fail` represent the total amount of failures 
+   - `module_name` represents the maven modules name
```
pjgg, line 14: Currently, the module name is the project root folder. The problem is that, If we have tree scenarios we will have tree moduleNames with the same name, and that is not very useful. What about replace `module_name` by `scenario_name` where scenario name is `rootFolderName_testClassName`

misc/Metrics.md
```diff
@@ -1,36 +1,34 @@
...
-   - `ts_quarkus_gauge_requests_total` represent the total amount of test
-   - `ts_quarkus_gauge_requests_success` represent the total amount of succeed test
-   - `ts_quarkus_gauge_requests_ignore` represent the total amount of ignored test
-   - `ts_quarkus_gauge_requests_fail` represent the total amount of failures 
-   - `ts_quarkus_gauge_modules` represents metrics associated to the test modules. You will have one of this metrics per quarkus scenario / module.
-        - `ts_quarkus_module_name` represents the maven modules name
-        - `ts_quarkus_module_status` could be `module_success` or `module_fail`
-   - `ts_quarkus_histogram_modules_duration`
+   - `tests_total` represent the total amount of test
+   - `tests_success` represent the total amount of succeed test
```
pjgg, line 11: what about... test_successed?

misc/Metrics.md
```diff
@@ -1,36 +1,34 @@
...
 - Metrics
-   - `ts_quarkus_gauge_requests_total` represent the total amount of test
-   - `ts_quarkus_gauge_requests_success` represent the total amount of succeed test
-   - `ts_quarkus_gauge_requests_ignore` represent the total amount of ignored test
-   - `ts_quarkus_gauge_requests_fail` represent the total amount of failures 
-   - `ts_quarkus_gauge_modules` represents metrics associated to the test modules. You will have one of this metrics per quarkus scenario / module.
-        - `ts_quarkus_module_name` represents the maven modules name
-        - `ts_quarkus_module_status` could be `module_success` or `module_fail`
-   - `ts_quarkus_histogram_modules_duration`
+   - `tests_total` represent the total amount of test
```
pjgg, line 10: tests_amount?

misc/Metrics.md
```diff
@@ -1,36 +1,34 @@
...
+   - `tests_success` represent the total amount of succeed test
+   - `tests_ignored` represent the total amount of ignored test
+   - `tests_fail` represent the total amount of failures
```
pjgg, line 13: tests_failed?

pjgg (changes requested):

misc/Metrics.md
Sgitario: +1

Sgitario: I kept the `module_name` and added a new one only for the scenario name `scenario_name` (for `testClassName`). 
Let me know if it looks better now.

misc/Metrics.md
Sgitario: +1

misc/Metrics.md
Sgitario: +1

misc/Metrics.md
Sgitario: I prefer either `total` or `count`, wdyt?

Sgitario:
@pjgg I've already addressed all your comments:
- `module_name` remains the same, but I've added an additional label `scenario_name` with the running class name
- renamed `module_duration_seconds` to `scenario_duration_seconds`
- renamed all the metrics from `success` to `succeed` and `fail` to `failed`

pjgg (approved):

PR #230, opened 2021-09-02
Move on to Kafka Quarkus apicurio registry

pjgg:
`quarkus-apicurio-registry-avro` was released and simplify the integration of Kafka + apicurio

The only dependency that now is required is the following one: 
```
 <dependency>
       <groupId>io.quarkus</groupId>
        <artifactId>quarkus-apicurio-registry-avro</artifactId>
 </dependency>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,13 +496,34 @@ public class StrimziKafkaWithRegistryMessagingIT {
 }
 ```
 
+The registry default values as `Docker image / version` and Registry API `path` could be overwritten by the following annotation properties:
+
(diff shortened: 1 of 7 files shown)


Sgitario (changes requested):
These changes are not compatible with 1.13 ++ the way you fixed the issue in `getSchemaRegistryUrl` is not correct.

Sgitario (approved):

Sgitario (changes requested):
I'm ok with the changes. The only missing bit is to update the README to explain that we can now overwrite the registry container.

Sgitario (approved):

PR #238, opened 2021-09-07
Compute all the properties when getting them using the Service

Sgitario:
Before, when doing `service.getProperty(X)`, this only returned the value from `service.withProperty(X)` regardless if the property X was defined in the `application.properties`.

Now, the service will return the final computed value from either the user entered or from the application properties or some custom properties added by the managed resource builders.


rsvoboda:
man, was just looking at it and it's merged

+1 for the change

Issue #237, opened 2021-09-06
Serverless deployments are not being fully deleted when not using ephemeral namespaces


Sgitario:
Related to https://github.com/quarkus-qe/quarkus-test-framework/pull/188

PR #239, opened 2021-09-07
Fix Serverless deployments ++ OpenShift deployments

Sgitario:
++ Minor refinements for Operators
++ Refactor OpenShift strategies handler to not use Reflection
++ Fix OpenShift exposed routes labels
++ Add labels and env vars properties for Knative deployments: This also fixes the issue to delete serverless resources when the ephemeral namespaces flag is disabled.

FIx https://github.com/quarkus-qe/quarkus-test-framework/issues/237


Sgitario:
run tests

Sgitario:
run tests

PR #253, opened 2021-09-13
Support Forced Dependencies

Sgitario:
### Forced Dependencies

We can also specify dependencies that are not part of the pom.xml by doing:

```java
@QuarkusScenario
public class GreetingResourceIT {

    private static final String HELLO = "Hello";
    private static final String HELLO_PATH = "/hello";
    private static final String INSTALLED_FEATURES_TEMPLATE = "Installed features: [cdi, %s, smallrye-context-propagation]";

    @QuarkusApplication(dependencies = @Dependency(artifactId = "quarkus-resteasy"))
    static final RestService blocking = new RestService();

    @QuarkusApplication(dependencies = @Dependency(artifactId = "quarkus-resteasy-reactive"))
    static final RestService reactive = new RestService();

    @Test
    public void shouldPickTheForcedDependencies() {
        // classic
        blocking.given().get(HELLO_PATH).then().body(is(HELLO));
        blocking.logs().assertContains(String.format(INSTALLED_FEATURES_TEMPLATE, "resteasy"));

        // reactive
        reactive.given().get(HELLO_PATH).then().body(is(HELLO));
        reactive.logs().assertContains(String.format(INSTALLED_FEATURES_TEMPLATE, "resteasy-reactive"));
    }
}
```

If no group ID and version provided, the framework will assume that the dependency is a Quarkus extension, so it will use the `quarkus.platform.groupId` (or `io.quarkus`) and the default Quarkus version.

This also can be used to append other dependencies apart from Quarkus.


Sgitario:
run tests

Sgitario:
run tests

Sgitario:
run tests

README.md
```diff
@@ -85,6 +85,41 @@ public class PingPongResourceIT {
...
+        // reactive
+        reactive.given().get(HELLO_PATH).then().body(is(HELLO));
+        reactive.logs().assertContains(String.format(INSTALLED_FEATURES_TEMPLATE, "resteasy-reactive"));
+    }
+}
+```
+
+If no group ID and version provided, the framework will assume that the dependency is a Quarkus extension, so it will use the `quarkus.platform.groupId` (or `io.quarkus`) and the default Quarkus version.
+
+This also can be used to append other dependencies apart from Quarkus.
```
rsvoboda, line 121: This expects full GAV to be specified, right?

examples/blocking-reactive-model/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        <version>0.0.10-SNAPSHOT</version>
+        <relativePath>../../</relativePath>
+    </parent>
+    <artifactId>examples-blocking-reactive</artifactId>
+    <name>Quarkus - Test Framework - Examples - Blocking + Reactive Model</name>
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy</artifactId>
+            <scope>provided</scope>
```
rsvoboda, line 16: So this is the trick to have code completition in IDE.

This could be documented as handy trick. 

You mentioned dev mode and remote dev mode won't work with this feature, should be also mentioned in readme

rsvoboda (changes requested):

README.md
Sgitario: No GAV, but separeted fields: one for groupId, artifactId and version.

examples/blocking-reactive-model/pom.xml
Sgitario: DevMode and RemoteDev uses a concrete annotation where the new field `dependencies` is not available, so I think we don't need to mention it.

About the handy trick, I will add the jax-rs dependency directory which I think it's more elegant, wdyt?

rsvoboda: Jax-rs api dependency ?

DevMode and RemoteDev ... ok, but I would still mention in README that "forced dependencies" are allowed for jvm / native mode, not for dev mode.

README.md
rsvoboda: that wasn't obvious to me, thanks.

Maybe it would be worth to enhance the example and add explicit groupId, 

```java
@QuarkusApplication(dependencies = @Dependency(artifactId = "quarkus-resteasy-reactive", groupId = "io.quarkus"))
```

Sgitario: Updated the README.

examples/blocking-reactive-model/pom.xml
Sgitario: I've already updated the README to state that this feature is not supported in Dev or Remote Dev scenarios.

I've also changed the example pom file to use the jax-rs api dependency instead of the quarkus-resteasy:
https://github.com/quarkus-qe/quarkus-test-framework/pull/253/files#diff-5f1afddf2b21f65a58e6aa212e01eb16621165ef2a238d221f8a98dc3f4b4f6bR15

mjurc (approved):

PR #258, opened 2021-09-14
Fix issue when escaped characters in properties file

Sgitario:
See issue https://stackoverflow.com/questions/10699055/how-do-you-escape-colon-in-properties-file.

Fixed by not using Properties.store method.


kshpak:
The issue is not fixed with '/' slashes still throwing:
Invalid value: "io.quarkus.ts.opentelemetry.PingPongService/mp-rest/scope": a valid environment variable name must consist of alphabetic characters, digits, '_', '-', or '.',

quarkus-test-core/src/main/java/io/quarkus/test/utils/PropertiesUtils.java
```diff
@@ -78,11 +78,13 @@ public static String resolveProperty(String value) {
...
-        try {
-            Properties properties = new Properties();
-            properties.putAll(map);
-            properties.store(new FileOutputStream(target.toAbsolutePath().toFile()), AUTOGENERATED_COMMENT);
+        Properties properties = new Properties();
+        properties.putAll(map);
+        try (PrintWriter pw = new PrintWriter(target.toFile())) {
+            pw.println(AUTOGENERATED_COMMENT);
+            for (Map.Entry<Object, Object> e : properties.entrySet()) {
+                pw.println(e.getKey() + "=" + e.getValue());
```
kshpak, line 86: Did you forget to close() pw?

kshpak: Ok, I see in try block is AutoClosable

PR #266, opened 2021-09-16
Delete service folder on close to free resources

Sgitario:


mjurc (approved):
Thanks for quick fix! 

LGTM.

Issue #240, opened 2021-09-07
Support of YAML configuration


Sgitario:
At the moment, the quarkus test framework is highly tied to properties file in the form of:

```
a.b=value
```

However, when using a YAML properties file for Quarkus, it might have some side effects (mostly when the test framework triggers a build operator to create the Quarkus applications at each scenario)

PR #281, opened 2021-09-28
Allow to use a custom application.properties per test scenario

Sgitario:
By default, the test framework will use the `application.properties` file at `src/main/resources` folder. The service interface provides multiple methods to add properties in the scenario:
- `service.withProperties(path)`
- `service.withProperty(key, value)`

If you want to use a different application properties file for all the tests, you can add the `application.properties` file at `src/test/resources` and the test framework will use this instead.

Moreover, if you want to select a concrete application properties file for a single test scenario, then you can configure your Quarkus application using:

```java
@QuarkusScenario
public class PingPongResourceIT {

    @QuarkusApplication(properties = "my-custom-properties.properties")
    static final RestService pingpong = new RestService();
}
```

This option is available also for Dev Mode, Remote Dev mode and remote git applications, and works for JVM, Native, OpenShift and Kubernetes. 

| Note that the test framework does not support the usage of YAML files yet [#240](https://github.com/quarkus-qe/quarkus-test-framework/issues/240)

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
@@ -53,6 +53,7 @@ public abstract class QuarkusApplicationManagedResourceBuilder implements Manage
     private List<AppDependency> forcedDependencies = Collections.emptyList();
     private boolean requiresCustomBuild = false;
     private ServiceContext context;
+    private String propertiesFile = APPLICATION_PROPERTIES;
     private boolean sslEnabled = false;
     private boolean grpcEnabled = false;
     private Map<String, String> propertiesSnapshot;
@@ -65,6 +66,10 @@ public abstract class QuarkusApplicationManagedResourceBuilder implements Manage
 
     protected void setContext(ServiceContext context) {
         this.context = context;
+    }
+
+    protected void setPropertiesFile(String propertiesFile) {
+        this.propertiesFile = propertiesFile;
     }
 
     protected boolean isSslEnabled() {
@@ -100,7 +105,7 @@ public abstract class QuarkusApplicationManagedResourceBuilder implements Manage
         Path applicationProperties = getComputedApplicationProperties();
         if (!Files.exists(applicationProperties)) {
             // computed properties have not been propagated yet, we use the one from src/main/resources
-            applicationProperties = RESOURCES_FOLDER.resolve(APPLICATION_PROPERTIES);
+            applicationProperties = RESOURCES_FOLDER.resolve(propertiesFile);
         }
 
         if (!Files.exists(applicationProperties)) {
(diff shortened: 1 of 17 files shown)


Sgitario:
run tests

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
```diff
@@ -53,6 +53,7 @@
     private List<AppDependency> forcedDependencies = Collections.emptyList();
     private boolean requiresCustomBuild = false;
```
pjgg, line 54: On this file, there is a "potential" null-pointer exception over line 48 (is not really related to this PR, was something that we left...)

`private static final Set<String> BUILD_PROPERTIES =  ... `

Could replace it with something more defensive as:

```
    private static final Set<String> BUILD_PROPERTIES = defaultBuildProperties();
.
.
.

 private static Set<String> defaultBuildProperties() {
        return defaultIfBlank(FileUtils.loadFile(BUILD_TIME_PROPERTIES), "").lines().collect(toSet());
    }
```

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
```diff
@@ -100,7 +105,7 @@ public String getComputedProperty(String name) {
         Path applicationProperties = getComputedApplicationProperties();
         if (!Files.exists(applicationProperties)) {
             // computed properties have not been propagated yet, we use the one from src/main/resources
-            applicationProperties = RESOURCES_FOLDER.resolve(APPLICATION_PROPERTIES);
+            applicationProperties = RESOURCES_FOLDER.resolve(propertiesFile);
         }
 
         if (!Files.exists(applicationProperties)) {
```
pjgg, line 111: Line 120 there is another potential "null pointer". My proposal is ament `PropertiesUtils.toMap`

```
public static Map<String, String> toMap(Path path) {
        Map<String, String> result = new HashMap<>();
        try (InputStream in = Files.newInputStream(path)) {
            result = toMap(in);
        } catch (IOException e) {
            fail("Could not load map from path. Caused by " + e);
        }

        return result;
    }
```

Again is not really related to this PR

pjgg: or maybe on `PropertiesUtils` you could have a helper function as `Map<String, String> getEmptyProperties()` that will return a `new HashMap<>();` sound "useless" but add semantic to the code and then later could be used in other places as here

```
 private void createComputedApplicationProperties() {
        Path sourceApplicationProperties = getResourcesApplicationFolder().resolve(propertiesFile);
        Path generatedApplicationProperties = getResourcesApplicationFolder().resolve(APPLICATION_PROPERTIES);
        Map<String, String> map = new HashMap<>();   // HERE
        // Add the content of the source application properties into the auto-generated application.properties
```

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
```diff
@@ -179,22 +184,23 @@ protected Path getResourcesApplicationFolder() {
...
 
     private void createComputedApplicationProperties() {
-        Path applicationProperties = getComputedApplicationProperties();
+        Path sourceApplicationProperties = getResourcesApplicationFolder().resolve(propertiesFile);
+        Path generatedApplicationProperties = getResourcesApplicationFolder().resolve(APPLICATION_PROPERTIES);
         Map<String, String> map = new HashMap<>();
-        // Put the original application properties
-        if (Files.exists(applicationProperties)) {
-            map.putAll(PropertiesUtils.toMap(applicationProperties));
+        // Add the content of the source application properties into the auto-generated application.properties
```
pjgg, line 195: I am not sure if this flow should have his own private methods
 - createFromSourceApplicationProperties
 - enrichWithServiceApplicationProperties
 - replaceMainApplicationProperties

pjgg (changes requested):
Over all looks good to me , just some cosmetic changes

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
Sgitario: Changed the FileUtils.loadFile to return empty which is never going to happen.

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
Sgitario: Changed the PropertiesUtils.toMap to return an empty map but is never going to happen.

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
Sgitario: This method has only 12 lines and the logic is not very complicated, I will rather keeping it like it is.

pjgg (approved):

PR #282, opened 2021-09-28
Fix File Logging for Scenarios

Sgitario:
Before, only the last scenario kept in the target/logs/tests.out file which made difficult to troubleshoot failures.

Now, all the scenarios will be logged in a separated log file and we will only keep only the log files when the scenario failed (test failures).


pjgg (approved):

pjgg:
LGTM, let's wait to CI

PR #291, opened 2021-09-30
Feature/hostnames

fedinskiy:
### The problem:
All classes, which we use for deploying services in our TS have two methods:
	getHost(returns something like "http://localhost")
	getPort(returns port as integer)
In most places, we call both these methods and then manually glue their output together, sometimes replacing http with other scheme.
In some cases, we have to parse output of getHost() string, since we need only real host name, without a scheme

### The solution:
Replace getHost and getPort with single method getURI, which will return an object of special class, with scheme(e.g. "http"), host(e.g. "localhost") and port(e.g. 8080) at the same time.
We will be able to pick parts of it manually, and glue them with single method.

UPD: initially, I planned to use java.net.URI, but it turns out, that kafka use schemes, which are incompatible with URI standard from IETF and, therefore with java implementation


rsvoboda:
What trouble/issue is this PR solving? There is no context in description of the PR nor in the change set.

fedinskiy:
@rsvoboda I have linked an issue, but for now It's just a draft, which I wanted to check on a server.

rsvoboda:
@fedinskiy how about this PR, is it still draft?

fedinskiy:
@rsvoboda from the technical perspective it's ready, but I would like a re-check it on the new version of TS. From the social perspective, there is a need to discus, whether we need these changes and whether everybody likes this approach.

PR #312, opened 2021-10-19
Update the README to keep it smaller with only the getting started

Sgitario:


mjurc:
The existing content looks good to me.

Do you think a link to wiki would be useful?

Sgitario:
> The existing content looks good to me.
> 
> Do you think a link to wiki would be useful?

There is a link to the wiki at the bottom of the page.

PR #338, opened 2021-11-16
Service log level implementation

pjgg:
- implement service log level filter
- reduce example/kafka log visibility to SEVERE

Feature example:

test.properties
```
ts.<YOUR SERVICE NAME>.log.level=FINE
```
External Doc about log level integer value(useful in order to understand the proposal):
 https://docs.oracle.com/javase/7/docs/api/java/util/logging/Level.html


mjurc:
@pjgg Wouldn't it be more prudent to keep some level of logging (info and higher like in standard runtime), or alternately to pipe the logs into a file rather than stdout?

pjgg:
@mjurc currently we are not supporting service log level
[Wiki log ref](https://github.com/quarkus-qe/quarkus-test-framework/wiki/2.-Configuration#logging)

Based on your feedback, I think that could be useful to implement it. By this way, we can set Kafka service as `SEVERE` or `WARNING` in order to log only those events.

mjurc:
LGTM, thank you.

PR #339, opened 2021-11-21
[#308] Update dockerfiles for OpenShift deployment

### LINKED ISSUE #308: Dockerfile.* are outdated, using old ubi image
https://github.com/quarkus-qe/quarkus-test-framework/tree/main/quarkus-test-images/src/main/resources

Using `FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3` instead of 8.4

There might be more inconsistences, please check with the newer versions of Dockerfile.* generated by code.quarkus.

mjurc:
* Updating dockerfiles used for OpenShift deployment to version used in
  Quarkus 2.4.

Fixes #308.

### DIFF
--- a/quarkus-test-images/src/main/resources/Dockerfile.jvm
+++ b/quarkus-test-images/src/main/resources/Dockerfile.jvm
@@ -1,4 +1,27 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3 
+####
+# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
+#
+# Before building the container image run:
+#
+# ./mvnw package
+#
+# Then, build the image with:
+#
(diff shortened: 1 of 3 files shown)


Sgitario (changes requested):
The property `ARTIFACT_PARENT` is used by the test framework to configure where the generated jar file by the test framework (for ex: target/MyTest/app/target/app.jar) is.
Therefore, in order to all the deployment strategies continue working, the change that the property `ARTIFACT_PARENT` is replaced by only `target` needs to be reverted.

mjurc:
> The property `ARTIFACT_PARENT` is used by the test framework to configure where the generated jar file by the test framework (for ex: target/MyTest/app/target/app.jar) is. Therefore, in order to all the deployment strategies continue working, the change that the property `ARTIFACT_PARENT` is replaced by only `target` needs to be reverted.

Good catch! I've updated the PR.

fedinskiy (approved):

PR #340, opened 2021-11-23
Downgrade confluent AVRO serializers to 6.2.1

pjgg:
Confluent AVRO serializers version 7.0 are supported in Upstream "999-SNAPSHOT" but TestFramework is pointing to Quarkus `2.3.0.Final` so we should downgrade Confluent serializer version from now. 

Confluent serializer 7.0 is also not supported in `Quarkus 2.5.0.Final` (Native mode)

Fix daily build: https://github.com/quarkus-qe/quarkus-test-framework/runs/4293313951?check_suite_focus=true

### DIFF
--- a/examples/kafka-registry/pom.xml
+++ b/examples/kafka-registry/pom.xml
@@ -10,7 +10,7 @@
     <artifactId>examples-kafka-registry</artifactId>
     <name>Quarkus - Test Framework - Examples - Kafka with Registry</name>
     <properties>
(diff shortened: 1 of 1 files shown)


rsvoboda (changes requested):
proper fix imho is to update Quarkus version to 2.5

pjgg:
> proper fix imho is to update Quarkus version to 2.5

doesn't get into 2.5 release

https://github.com/quarkusio/quarkus/issues/21412

Looks that is planned to 2.6

rsvoboda (approved):

PR #344, opened 2021-11-24
`delete project after all` property should not scale down services

### LINKED ISSUE #342: Feature: Don't scale down OpenShift resources if required
To be able to investigate what happened with OpenShift scenario and its services it will be handy to leave resources untouched after the test has been executed (e.x. if the test failed and we want to see pod logs)

Currently, we have only property `-Dts.openshift.delete.project.after.all=false` which won't remove resources, however, it will scale them down.

Implementation ideas: the new property can be introduced which will be used along with `-Dts.openshift.delete.project.after.all=false`. Because some modules contain more than one OpenShift test we can't bind property `-Dts.openshift.delete.project.after.all=false` to do both (not delete + not scale down) (I guess)

pjgg:
`ts.kubernetes.delete.project.after.all=false` should not scale down k8s/ocp services
Also must only works if `ts.kubernetes.ephemeral.namespaces.enabled=false`

Close #342

### DIFF
--- a/quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
+++ b/quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
@@ -1,5 +1,7 @@
 package io.quarkus.test.bootstrap;
 
+import static io.quarkus.test.bootstrap.inject.KubectlClient.ENABLED_EPHEMERAL_NAMESPACES;
+
 import java.nio.file.Path;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -21,11 +23,21 @@ public class KubernetesExtensionBootstrap implements ExtensionBootstrap {
 
     @Override
     public boolean appliesFor(ScenarioContext context) {
-        return context.isAnnotationPresent(KubernetesScenario.class);
+        boolean isValidConfig = context.isAnnotationPresent(KubernetesScenario.class);
+        if (isValidConfig && !DELETE_NAMESPACE_AFTER.getAsBoolean() && ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean()) {
+            Log.error("-Dts.kubernetes.delete.project.after.all=false is only supported with"
+                    + " -Dts.kubernetes.ephemeral.namespaces.enabled=false");
+            isValidConfig = false;
+        }
+
+        return isValidConfig;
     }
 
     @Override
     public void beforeAll(ScenarioContext context) {
+        // if deleteNamespace and ephemeral namespaces are disabled then we are in debug mode. This mode is going to keep
+        // all scenario resources in order to allow you to debug by yourself
+        context.setDebug(!DELETE_NAMESPACE_AFTER.getAsBoolean() && !ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean());
         client = KubectlClient.create(context.getId());
     }
 
--- a/quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/OpenShiftExtensionBootstrap.java
+++ b/quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/OpenShiftExtensionBootstrap.java
@@ -1,5 +1,7 @@
 package io.quarkus.test.bootstrap;
 
+import static io.quarkus.test.bootstrap.inject.OpenShiftClient.ENABLED_EPHEMERAL_NAMESPACES;
+
 import java.nio.file.Path;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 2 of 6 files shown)


pjgg:
Pinpong serverless scenario `OpenShiftUsingExtensionAndDockerBuildAndServerlessPingPongResourceIT` is failling by an error not related to this PR 

```
SEVERE Caught exception while closing extension context: org.junit.jupiter.engine.descriptor.ClassExtensionContext@14823f76: javax.net.ssl.SSLException: Unsupported or unrecognized SSL message
```

Sgitario (changes requested):
If you add this code in the stop methods, then when calling the stop method in a test scenario, it wont work. The user would expect the app is stopped but it wont because the ephemeral was enabled...

Therefore, this is not a suitable fix.

Moreover, in k8s there are now two properties: "ts.kubernetes.delete.project.after.all" (this was created here and "ts.kubernetes.delete.namespace.after.all" (the good one as k8s does not know what a project is - this concept is from ocp)

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
```diff
@@ -16,16 +16,28 @@
 
     private static final PropertyLookup DELETE_NAMESPACE_AFTER = new PropertyLookup("ts.kubernetes.delete.namespace.after.all",
             Boolean.TRUE.toString());
+    private static final PropertyLookup DELETE_PROJECT_AFTER = new PropertyLookup("ts.kubernetes.delete.project.after.all",
```
Sgitario, line 19: This property does not make sense in k8s, we should use `DELETE_NAMESPACE_AFTER` instead.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
```diff
@@ -16,16 +16,28 @@
...
+    private static final PropertyLookup ENABLED_EPHEMERAL_NAMESPACES = new PropertyLookup(
+            "ts.kubernetes.ephemeral.namespaces.enabled", Boolean.TRUE.toString());
 
     private KubectlClient client;
 
     @Override
     public boolean appliesFor(ScenarioContext context) {
-        return context.isAnnotationPresent(KubernetesScenario.class);
+        boolean isValidConfig = context.isAnnotationPresent(KubernetesScenario.class);
+        if (isValidConfig && !DELETE_PROJECT_AFTER.getAsBoolean() && ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean()) {
```
Sgitario, line 29: This logic is wrong, if the KubernetesScenario annotation is not present, we should not check those properties.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
```diff
@@ -16,16 +16,28 @@
...
+                    + " -Dts.kubernetes.ephemeral.namespaces.enabled=false");
+            isValidConfig = false;
+        }
+
+        return isValidConfig;
     }
 
     @Override
     public void beforeAll(ScenarioContext context) {
+        context.setDebug(!DELETE_PROJECT_AFTER.getAsBoolean() && !ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean());
```
Sgitario, line 40: nitpick comment: I would add a comment explaining your purpose here: explain better why it should be set DEBUG if these two properties are set.

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
```diff
@@ -16,16 +16,28 @@
 
     private static final PropertyLookup DELETE_NAMESPACE_AFTER = new PropertyLookup("ts.kubernetes.delete.namespace.after.all",
             Boolean.TRUE.toString());
+    private static final PropertyLookup DELETE_PROJECT_AFTER = new PropertyLookup("ts.kubernetes.delete.project.after.all",
+            Boolean.TRUE.toString());
+    private static final PropertyLookup ENABLED_EPHEMERAL_NAMESPACES = new PropertyLookup(
```
Sgitario, line 21: This property definition is duplicated in KubectlClient

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/OpenShiftExtensionBootstrap.java
```diff
@@ -20,16 +20,26 @@
             Boolean.TRUE.toString());
     private static final PropertyLookup DELETE_PROJECT_AFTER = new PropertyLookup("ts.openshift.delete.project.after.all",
             Boolean.TRUE.toString());
+    private static final PropertyLookup ENABLED_EPHEMERAL_NAMESPACES = new PropertyLookup(
```
Sgitario, line 23: This property definition is duplicated in OpenShiftClient.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/OpenShiftExtensionBootstrap.java
```diff
@@ -20,16 +20,26 @@
...
+            "ts.openshift.ephemeral.namespaces.enabled", Boolean.TRUE.toString());
 
     private OpenShiftClient client;
 
     @Override
     public boolean appliesFor(ScenarioContext context) {
-        return context.isAnnotationPresent(OpenShiftScenario.class);
+        boolean isValidConfig = context.isAnnotationPresent(OpenShiftScenario.class);
+        if (isValidConfig && !DELETE_PROJECT_AFTER.getAsBoolean() && ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean()) {
```
Sgitario, line 31: This logic is wrong, if the OpenShiftScenario annotation is not present, we should not check those properties.

quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/OpenShiftExtensionBootstrap.java
```diff
@@ -20,16 +20,26 @@
...
+                    + " -Dts.openshift.ephemeral.namespaces.enabled=false");
+            isValidConfig = false;
+        }
+
+        return isValidConfig;
     }
 
     @Override
     public void beforeAll(ScenarioContext context) {
+        context.setDebug(!DELETE_PROJECT_AFTER.getAsBoolean() && !ENABLED_EPHEMERAL_NAMESPACES.getAsBoolean());
```
Sgitario, line 42: nitpick comment: I would add a comment explaining your purpose here: explain better why it should be set DEBUG if these two properties are set.

Sgitario (changes requested):

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/KubernetesExtensionBootstrap.java
Sgitario: I read this wrongly. Deleted.

Sgitario (approved):

PR #346, opened 2021-11-26
Make containers reusable

DavideD:
Close https://github.com/quarkus-qe/quarkus-test-framework/issues/347

This is a POC of what I had in mind when I mentioned that it should be possible to make containers reusable.
But, ideally, instead of the annotation, it would be more powerful to set this with a `DatabaseService.withReuse(true)` but I couldn't figure out how to do it

### DIFF
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/Container.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/Container.java
@@ -20,4 +20,6 @@ public @interface Container {
     String[] command() default {};
 
     Class<? extends ManagedResourceBuilder> builder() default ContainerManagedResourceBuilder.class;
+
+    boolean reusable() default false;
 }
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/ContainerManagedResourceBuilder.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/ContainerManagedResourceBuilder.java
@@ -19,6 +19,7 @@ public class ContainerManagedResourceBuilder implements ManagedResourceBuilder {
     private String expectedLog;
     private String[] command;
     private Integer port;
+    private boolean reusable;
 
     protected String getImage() {
         return image;
@@ -40,6 +41,10 @@ public class ContainerManagedResourceBuilder implements ManagedResourceBuilder {
         return context;
     }
 
+    public boolean isReusable() {
+        return reusable;
+    }
+
     @Override
     public void init(Annotation annotation) {
         Container metadata = (Container) annotation;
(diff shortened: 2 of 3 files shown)


pjgg:
Thank you @DavideD instead of an annotation let's do it with a property that could be set in `test.properties` file as currently is done with `privilege mode`.

So, I think that the solution should look almost the same as [privilege mode](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java#L11) implementation 

I think that first, we should create a test scenario in order to prove our solution. You could use any of the examples that comes on this repository, for example [database-mysql](https://github.com/quarkus-qe/quarkus-test-framework/tree/main/examples/database-mysql)

Create an Scenario `MySqlReusableDatabaseIT` that should look almost the same as `MySqlDatabaseIT`. Is important that both scenario has the same `MySqlService` service name, for example, we could named `database`

MySqlDatabaseIT.java
```
@QuarkusScenario
public class MySqlDatabaseIT extends AbstractSqlDatabaseIT {

    static final int MYSQL_PORT = 3306;

    @Container(image = "quay.io/bitnami/mysql:8.0.27", port = MYSQL_PORT, expectedLog = "ready for connections")
    static MySqlService database = new MySqlService();

.
.
.
.
```

MySqlReusableDatabaseIT
```
@QuarkusScenario
public class MySqlReusableDatabaseIT extends AbstractSqlDatabaseIT {

    static final int MYSQL_PORT = 3306;

    @Container(image = "quay.io/bitnami/mysql:8.0.27", port = MYSQL_PORT, expectedLog = "ready for connections")
    static MySqlService database = new MySqlService();

.
.
.

```

Then lets add a this property `ts.database.container.reusable=true` into `test.properties`
If for example, we have a project with several databases and we only want to share `db2` instances, then DB2Services should be named as `mydb2instance` (for example) or just db2, and then in your `test.property` file add the property `ts.mydb2instance.container.reusable=true`

Then we could proceed with the implementation as  [privilege mode](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java#L11) 

We should create a new property called:
```
private static final String REUSABLE_MODE = "container.reusable";
```
and then in `initContainer()` do the business logic `container.withReuse(true);` only if isReusable. 

And don't miss this method too (as privilege mode): 
```
protected boolean isReusable() {
        return model.getContext().getOwner().getConfiguration().isTrue(REUSABLE_MODE);
    }
```

DavideD:
Sounds like it would work but I don't have the time at the moment to work on this.

pjgg:
Driven by: https://github.com/quarkus-qe/quarkus-test-framework/pull/349

PR #349, opened 2021-11-30
Reuse a docker containers between Quarkus-scenarios

### LINKED ISSUE #347: Reuse a docker containers between Quarkus-scenarios
Currently, we are sharing containers between tests of the same scenario. We would like to do the same but between Scenarios, for those containers that are too heavy as DB2. 

Reusable containers should be configurable by properties as nowadays we are doing with privilege mode(test. properties)
For example:

```
ts.db2.container.privileged-mode=true
ts.db2.container.reusable=true
```

pjgg:
close #347

depends on: https://github.com/quarkus-qe/quarkus-test-framework/pull/350

### DIFF
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java
@@ -10,6 +10,7 @@ import io.quarkus.test.utils.DockerUtils;
 public class GenericDockerContainerManagedResource extends DockerContainerManagedResource {
 
     private static final String PRIVILEGED_MODE = "container.privileged-mode";
+    private static final String REUSABLE_MODE = "container.reusable";
     private final ContainerManagedResourceBuilder model;
 
     protected GenericDockerContainerManagedResource(ContainerManagedResourceBuilder model) {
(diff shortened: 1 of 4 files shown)


quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java
```diff
@@ -43,11 +44,20 @@ protected int getTargetPort() {
             container.setPrivilegedMode(true);
         }
 
+        if (isReusable()) {
+            Log.info(model.getContext().getOwner(), "Running container on Reusable mode");
+            container.withReuse(true);
```
mjurc, line 49: I would rather postpone this as this is marked as unstable API in testcontainers

mjurc (commented):
I'm not quite sure what to do with this. I am not really comfortable with merging into framework something that includes unstable upstream API.

What are the actual gains we're expecting?

quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java
pjgg: DocRef: https://javadocs.dev/org.testcontainers/testcontainers/1.16.0/org/testcontainers/containers/GenericContainer.html#withReuse-boolean-

Yes,  is totally true that is marked as `@UnstableAPI` that means: "Marks that the annotated API is a subject to change and SHOULD NOT be considered a stable API."

I will not use it internally in Quarkus test suite BUT other teams request it. In my opinion, it doesn't hurt. I have added a warning log in order to warm to the end-user about this `UnstableAPI`

pjgg: Also on the wiki we could emphasize more that is not a stable API, or that is using test containers "withReuse" method that is UnstableAPI

rsvoboda:
I think we can have this feature as it was asked by people outside the team, it's not enabled by default, so no need to be too worried now.

If this feature gets used in TS there is one drawback to consider: it can hurt us when something goes sideways in any of the previous test, searching which test caused an side effect can be a pity! I'd rather have longer execution time, our time is more precious :)

Details from Pablo:
```
this PR come from a conversation in Zulip
with Sanne and DavideD
https://github.com/quarkus-qe/quarkus-test-framework/pull/346

basically for them is a pain, because DB2 takes so much time to start
so if they are testing different Quarkus config over DB2
they would like to reuse his DB2 instance
```

pjgg:
Rebased against the latest main version

pjgg:
Close: #349

mjurc (approved):
LGTM

PR #350, opened 2021-11-30
Example/microprofile required grpc-api lib on their classpath

pjgg:
gRPC metrics are now handled by the quarkus-micrometer extension. gRPC server and clients metrics are automatically enabled when your application depends on the quarkus-micrometer extension. In the case you do not want to collect these metrics, you can disable them using:
```
quarkus.micrometer.binder.grpc-client.enabled=false
quarkus.micrometer.binder.grpc-server.enabled=false
```

Doc Ref: https://github.com/quarkusio/quarkus/wiki/Migration-Guide-2.6#grpc-metrics-changes

### DIFF
--- a/examples/microprofile/src/main/resources/application.properties
+++ b/examples/microprofile/src/main/resources/application.properties
@@ -1,1 +1,6 @@
+# GRPC metrics are enabled by default if you bring "grpc-api" as a dependency and "quarkus-micrometer".
+# In our case "quarkus-test-core" brings it.
+# //TODO https://github.com/quarkus-qe/quarkus-test-framework/issues/359
 
+quarkus.micrometer.binder.grpc-client.enabled=false
+quarkus.micrometer.binder.grpc-server.enabled=false


pjgg:
ping pong issue is not related to this PR.

rsvoboda:
Some additional context in https://github.com/quarkusio/quarkus/pull/21654#issuecomment-979262107

rsvoboda (changes requested):
this exampleis just jax-rs, nothing else

The app has nothing to do with grpc and I still need to disable metrics for it ?
Why is need for grpc triggered at all ?

Please investigate more - check dep tree if the framework is not pulling in some grps deps / create standalone app without our framework there

examples/microprofile/src/main/resources/application.properties
```diff
@@ -1 +1,6 @@
+# GRPC metrics are enabled by default if you bring "grpc-api" as a dependency and "quarkus-micrometer".
+# In our case "quarkus-test-core" bring it.
```
rsvoboda, line 2:
```suggestion
# In our case "quarkus-test-core" brings it.
```

rsvoboda (changes requested):

rsvoboda (approved):

PR #355, opened 2021-12-14
Include relocations in Quarkus main build

jsmrcka:
In order to make use of maven relocations introduced by
https://github.com/quarkusio/quarkus/pull/21736, the `relocations`
profile is added to GH jobs that perform build of Quarkus main.

See also https://github.com/quarkusio/quarkus/wiki/Migration-Guide-2.6.

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -28,7 +28,7 @@ jobs:
           check-latest: true
       - name: Build Quarkus main
         run: |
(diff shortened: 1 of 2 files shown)


jsmrcka:
This should resolve failures in `examples/amq-tcp` and `examples/consul`.

There is another issue in `examples/microprofile`, which is unrelated.

fedinskiy (approved):

PR #358, opened 2021-12-16
Docker containers prefix feature

pjgg:
The goal of this PR is to provide the ability to set a custom docker container process prefix name. This feature will give us the ability to share a remote docker-engine between different Jenkins workers. So, if a Jenkins worker removes or makes some change to a docker container instance, will only apply to his "domain" or prefix. Basically is a workaround, is like a namespace based on prefixes.  Other technologies as Redis are using this prefix approach and works pretty well. 

The docker container names will be generated by our `DockerUtils class` and are going to be a random number plus an optional custom prefix.             

For example:

Imagine that you have two Jenkins jobs called `2.2-baremetal-ts-jvm` and `1.7-baremetal-ts-jvm`. Lets hash and then pick the first 10 digits, in order to have a unique name without weird characters:

`2.2-baremetal-ts-jvm` -> md5 -> `4a73c71664d6629d8fd639781b8a1b54` -> `4a73c71664`
`1.7-baremetal-ts-jvm` -> md5 -> `ba40ab410ed194be244b7114e2229d15` -> `ba40ab410e`

Then both jobs could be triggered with a command like the following one: 

`2.2-baremetal-ts-jvm` -> `mvn clean verify -Dts.global.docker-container-prefix=4a73c71664`
`1.7-baremetal-ts-jvm` -> `mvn clean verify -Dts.global.docker-container-prefix=ba40ab410e`

the result would be something like this (assuming that both jobs creates a Keycloak container): 

```
CONTAINER ID   IMAGE                               PORTS                                                   NAMES
e7a2ef3a3e84   quay.io/keycloak/keycloak:14.0.0    8443/tcp, 0.0.0.0:50138->8080/tcp, :::50138->8080/tcp   4a73c71664-2074505752

a6b3bf6c3c99   quay.io/keycloak/keycloak:14.0.0    8443/tcp, 0.0.0.0:50138->8080/tcp, :::50138->8080/tcp   ba40ab410e-2074505752
```

This prefix approach could be handly because allow us to then later clean the containers by running a script, something that is very interesting when you are using a remote docker engine between several workers:

Stop&remove all `4a73c71664` (2.2-baremetal-ts-jvm) containers:

```
docker stop $(docker ps -a | grep 4a73c71664 | awk '{print $1}')
docker rm $(docker ps -a | grep 4a73c71664 | awk '{print $1}')
```

Or if you want to stop and remove all containers of all jobs except Docker-api proxy then you could do something like:

```
docker stop $(docker ps -a | grep -v "docker-remote-api" | awk 'NR>1 {print $1}')
docker rm $(docker ps -a | grep -v "docker-remote-api" | awk 'NR>1 {print $1}')
```

### DIFF
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/GenericDockerContainerManagedResource.java
@@ -5,6 +5,7 @@ import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
 
 import io.quarkus.test.logging.Log;
+import io.quarkus.test.utils.DockerUtils;
 
 public class GenericDockerContainerManagedResource extends DockerContainerManagedResource {
 
@@ -43,6 +44,7 @@ public class GenericDockerContainerManagedResource extends DockerContainerManage
             container.setPrivilegedMode(true);
         }
 
+        container.withCreateContainerCmdModifier(cmd -> cmd.withName(DockerUtils.generateDockerContainerName()));
         container.withExposedPorts(model.getPort());
 
         return container;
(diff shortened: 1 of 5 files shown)


pjgg:
Baremetal microprofile error and OCP pingpong issue is not related to this PR

rsvoboda (changes requested):
Code looks fine to me.

I miss documentation PR for wiki update

pjgg:
> Code looks fine to me.
> 
> I miss documentation PR for wiki update

https://github.com/quarkus-qe/quarkus-test-framework/wiki/5.-Containers#custom-prefix-name

rsvoboda (approved):

PR #392, opened 2022-02-03
[fix] Disable also ColorPatternFormatter when property is used

llowinge:


rsvoboda (approved):

rsvoboda:
Thank you for the contribution!

PR #402, opened 2022-02-09
Set Quarkus_maven_plugin based on -Dquarkus-plugin.version parameter

pjgg:
If no quarkus plugin version is provided, by mvn command system property `-Dquarkus-plugin.version` then quarkus plugin version will have the same value as quarkus platform version. Otherwise  `-Dquarkus-plugin.version` value will be taken in place.

Examples:

`mvn clean verify`

quarkus.platform.version -> 2.2.5.Final
quarkus-plugin.version -> 2.2.5.Final

`mvn clean verify  -Dquarkus.platform.group-id=com.redhat.quarkus.platform -Dquarkus.platform.artifact-id=quarkus-bom -Dquarkus.platform.version=2.2.5.Final-redhat-00006 -Dquarkus-plugin.version=2.2.5.Final-redhat-00009`

quarkus.platform.version -> 2.2.5.Final-redhat-00006
quarkus-plugin.version -> 2.2.5.Final-redhat-00009

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/GitRepositoryQuarkusApplication.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/GitRepositoryQuarkusApplication.java
@@ -14,7 +14,9 @@ public @interface GitRepositoryQuarkusApplication {
 
     String contextDir() default "";
 
-    String mavenArgs() default "-DskipTests=true -DskipITs=true -Dquarkus.platform.version=${QUARKUS_VERSION}";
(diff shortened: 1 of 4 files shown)


rsvoboda (approved):
I don't like QUARKUS_, QUARKUS-, quarkus-plugin.version vs quarkus.plugin.version .. who can see the difference.

We should get to better naming, and some info should be added into comments to provide some context for the future when new people look into the code.

This can be done a bit later, but please create GH issue for that.

PR #428, opened 2022-03-10
Implement reactive URL in `OracleService`

jsmrcka:
Based on Reactive Oracle Client implementation:
https://github.com/quarkusio/quarkus/pull/22810


mjurc (changes requested):
This seems legitimate - but can you add a reactive Oracle test to `examples`? Just so that we have test coverage for this feature on framework side too.

rsvoboda:
Separate issue was created for the example app. Going to merge this change.

rsvoboda (approved):

PR #450, opened 2022-04-14
[fix] Wait for DC ready before scaling

llowinge:
It can happen, that the DC is not ready, in such case is better to wait some time.


llowinge:
@mjurc Can you look ?

mjurc (approved):
LGTM, the failure in CLI tests was unreleated, let's wait for the CI for green

rsvoboda:
@llowinge this PR causes troubles in serverless use-case. @mjurc will prepare fix for that.

Looking at the description of this PR I miss more context - what is the problem which the PR fixes / in which cases is not DeploymentConfig available? `It can happen` doesn't feel (to me) like a sufficient explanation

llowinge:
@rsvoboda It happened to me when OpenShift wasn't fast/stable probably and DC wasn't yet created, because after this change i didn't see it again (i mean eg. seeing `Waiting for dc to be ready` multiple times atleast). If it causes troubles you can revert it and if i hit problem in future i will do proper analysis what is wrong. Sorry for inconvenience.

rsvoboda:
The problem is fixed now, Michal did the PR yesterday.

OpenShift wasn't fast/stable ... OCP running in PSI? or local OKD instance?

llowinge:
OCP running in PSI.

PR #516, opened 2022-07-15
Decouple Jaeger from Test core by using OpenTelemetry

michalvavrik:
### Summary

`io.jaegertracing:jaeger-client` brings transitively forbidden dependencies we want to avoid. This PR replaces Jaeger client with an OpenTelemetry one as OpenTelemetry is suppported option.

Breaking changes: tracing newly uses `OpenTelemetry Protocol (OTLP) over gRPC` on port `4317` (protocol `HTTP`), and `jaegertracing/all-in-one` in version 1.35+ is required. `ts.global.tracing.jaeger.endpoint` should consist only of a base url (drop `/api/traces`). We need follow-up PR for Jenkins jobs and [upgrade](https://www.jaegertracing.io/docs/1.36/apis/#opentelemetry-protocol-stable) production Jaeger (as its version is 1.30).

I tried to keep as close as possible to the pre-PR state by following this [OpenTelemetry to Jaeger Transformation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md).

- [ ] Dependency update
- [x] Refactoring

### Checklist:


extensions/tracing/src/test/resources/test.properties
```diff
@@ -1 +1 @@
-ts.global.tracing.jaeger.endpoint=http://localhost:14268/api/traces
+ts.global.tracing.jaeger.endpoint=http://localhost:4317/api/traces
```
pjgg, line 1: `/api/traces` is still valid?

quarkus-test-core/src/main/java/io/quarkus/test/tracing/QuarkusScenarioAttributes.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import io.opentelemetry.api.trace.Span;
+import io.quarkus.test.utils.TestExecutionProperties;
+
+public class QuarkusScenarioAttributes {
+    public static final String SUCCESS = "success";
+    public static final String ERROR = "error";
```
pjgg, line 12: You ar not using this property at all

quarkus-test-core/src/main/java/io/quarkus/test/tracing/TracingExtensionBootstrap.java
```diff
@@ -19,11 +17,7 @@ public class TracingExtensionBootstrap implements ExtensionBootstrap {
     public TracingExtensionBootstrap() {
         String jaegerHttpEndpoint = JAEGER_HTTP_ENDPOINT_SYSTEM_PROPERTY.get();
         if (StringUtils.isNotEmpty(jaegerHttpEndpoint)) {
-            try {
-                quarkusScenarioTracer = new QuarkusScenarioTracer(jaegerHttpEndpoint);
-            } catch (TTransportException e) {
-                Log.error("Error setting up tracing capabilities. Turning off the tracing. Caused by: " + e.getMessage());
-            }
+            quarkusScenarioTracer = new QuarkusScenarioTracer(jaegerHttpEndpoint);
```
pjgg, line 20: If something goes wrong we want to log the error but continue with the test execution, because scenario tracing is not an stopper. Did you test a `jaegertracing/all-in-one:1.36` fatal error and see if the scenarios continue the regular execution?

pjgg (changes requested):

quarkus-test-core/src/main/java/io/quarkus/test/tracing/QuarkusScenarioSpan.java
```diff
@@ -29,11 +32,15 @@ public Span getOrCreate(ScenarioContext context) {
         return spanBucket.get(operationName);
     }
 
-    public Span save(Map<String, ?> logs, Set<String> tags, ScenarioContext scenarioContext) {
-        String operationName = getOperationName(scenarioContext);
-        Span span = spanBucket.get(operationName);
-        span.log(logs);
-        tags.forEach(tag -> span.setTag(tag, true));
+    public Span save(Set<String> attributes, ScenarioContext scenarioContext, Throwable throwable) {
```
pjgg, line 35: Maybe `attributes` could be a Map<String,boolean>` (or you could overwrite this method). This signature will allow us to set the Success attribute to false, onError event.

pjgg (changes requested):

quarkus-test-core/src/main/java/io/quarkus/test/tracing/QuarkusScenarioTracer.java
```diff
@@ -1,79 +1,68 @@
...
     }
 
     public void finishWithError(ScenarioContext context, Throwable cause) {
-        finishWithError(context, cause, ERROR);
-    }
-
-    public void finishWithError(ScenarioContext context, Throwable cause, String tag) {
-        Map<String, ?> err = Map.of(Fields.EVENT, "error", Fields.ERROR_OBJECT, cause, Fields.MESSAGE, cause.getMessage());
-        quarkusScenarioSpan.save(err, newHashSet(tag), context).finish();
+        quarkusScenarioSpan.save(Set.of(), context, cause).end();
```
pjgg, line 50: Maybe onError, we could set a success attribute to false
`success=false`

extensions/tracing/src/test/resources/test.properties
michalvavrik: hmm, looks like path is irrelevant here as `okhttp3.HttpUrl.Builder#parse` only takes a `host` and `port`, however according to the manual [only host and port](https://opentelemetry.io/docs/instrumentation/java/manual/#exporter) should be set, so I'll fix it. Good catch, thanks.

quarkus-test-core/src/main/java/io/quarkus/test/tracing/TracingExtensionBootstrap.java
michalvavrik: I took me quite a while to correctly set up Jaeger and when endpoint is both unavailable (f.e. not running)/returns unimplemented (f.e. wrong port), similar message is logged `SEVERE Failed to export spans. The request could not be executed. Full error message: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4317` and testing continues unaffected. I can't pledge for all possible runtime exceptions, but there are no checked and examples on OpenTelemetry reference do no catching (though failure might be blocker for them...). I could only catch `Exception` there, but I don't think it's a good practice.

quarkus-test-core/src/main/java/io/quarkus/test/tracing/QuarkusScenarioAttributes.java
michalvavrik: Jaeger adapter is doing this for us, please have a look https://github.com/open-telemetry/opentelemetry-java/blob/d42c2629b6fb3d4ce1e511454c8e853403cecec7/exporters/jaeger/src/main/java/io/opentelemetry/exporters/jaeger/Adapter.java#L116

michalvavrik: So I'll drop it.

michalvavrik:
Thank you for the review.

quarkus-test-core/src/main/java/io/quarkus/test/tracing/TracingExtensionBootstrap.java
pjgg: This is enough `EVERE Failed to export spans.` thanks!

pjgg (approved):
LGTM

Issue #263, opened 2021-09-14
BaseService cannot return hostname without protocol


fedinskiy:
BaseService and its descendant classes, e.g. RestService, have method getHost, which returns string, which looks like "http://localhost". In most places our code either replaces "http" in this string with some other protocol(e.g. "jdbc") or glues it together with results of getPort method of the same class. Also, there are problems with API, which accepts only URLs or URIs, since we have to parse output of those methods.

I suggest to either:
a) Replace these two methods with method getURL (or getURI)
b) Change the method getHost, so it will return only a host.

Sgitario:
+1 this area needs a lot of improvement. Feel free to refactor the solution or ideas.

PR #535, opened 2022-08-16
Using a special class for passing service URIs

fedinskiy:
### The problem:
All classes, which we use for deploying services in our TS have two methods:
	getHost(returns something like "http://localhost")
	getPort(returns port as integer)
In most places, we call both these methods and then manually glue their output together, sometimes replacing http with other scheme.
In some cases, we have to parse output of getHost() string, since we need only real host name, without a scheme

### The solution:
Replace getHost and getPort with single method getURI, which will return an object of special class, with scheme(e.g. "http"), host(e.g. "localhost") and port(e.g. 8080) at the same time.
We will be able to pick parts of it manually, and glue them with single method.

Initially, I planned to use java.net.URI, but it turns out, that kafka use schemes, which are incompatible with URI standard from IETF and, therefore with java implementation

Closes https://github.com/quarkus-qe/quarkus-test-framework/issues/263

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/URILike.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/URILike.java
@@ -1,0 +1,159 @@
+package io.quarkus.test.services;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class URILike {
+
+    private static final String SCHEME_SEPARATOR = ":";
+    private static final String HOST_PREFIX = "//";
+    /*
+     * Things like "SASL_PLAINTEXT" are not valid schemes from the URI point of view,
+     * but we need to process them anyway
+     */
+    private final String scheme;
+    private final URI wrapped;
+
+    public URILike(String scheme, String userinfo, String host, int port, String path) {
+        this.scheme = scheme;
+        this.wrapped = createURI(
+                null, //we process scheme separately
+                userinfo,
+                host,
+                port,
+                path,
+                null, // query is not used at the time of writing
+                null // fragment is not used at the time of writing
+        );
+    }
+
+    public URILike(String scheme, String host, int port, String path) {
+        this(scheme, null, host, port, path);
+    }
+
+    private URILike(String scheme, URI uri) {
+        this.scheme = scheme;
+        this.wrapped = uri;
+    }
+
+    /**
+     * Fluently creates new object, but replaces "scheme" part.
+     *
+     * @param scheme — String, will be used as a "scheme" part in the resulting string.
+     *        May not be conforming to https://datatracker.ietf.org/doc/html/rfc3986/#section-3.1
+     * @return new object with changed value of "scheme"
+     */
+    public URILike withScheme(String scheme) {
+        return new URILike(scheme, this.wrapped);
+    }
+
+    public URILike withPath(String path) {
+        URI wrapped = this.wrapped;
+        var withPath = createURI(wrapped.getScheme(),
+                wrapped.getUserInfo(),
+                wrapped.getHost(),
+                wrapped.getPort(),
+                path,
+                wrapped.getQuery(),
+                wrapped.getFragment());
+        return new URILike(this.scheme, withPath);
+    }
+
+    public URILike withPort(int port) {
+        URI wrapped = this.wrapped;
+        var withPath = createURI(wrapped.getScheme(),
+                wrapped.getUserInfo(),
+                wrapped.getHost(),
+                port,
+                wrapped.getPath(),
+                wrapped.getQuery(),
+                wrapped.getFragment());
+        return new URILike(this.scheme, withPath);
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        if (scheme != null) {
+            sb.append(scheme);
+            sb.append(SCHEME_SEPARATOR);
+        }
+        sb.append(wrapped);
+        String result = sb.toString();
+        if (result.startsWith("//")) {
+            return result.substring(2);
+        } else {
+            return result;
+        }
+    }
+
+    public String getScheme() {
+        return this.scheme;
+    }
+
+    public String getHost() {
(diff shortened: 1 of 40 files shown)


pjgg (changes requested):
Let's try to keep `getHost` and `getPort` method as `@deprecated` and be backward compatible for a couple of releases.

fedinskiy:
run tests

fedinskiy:
Fails in Openshift Native are due to OOM/Out of Space during the build, and they do not reproduce themselves in a local run. This PR is ready for a check

examples/jaeger/src/test/java/io/quarkus/qe/ClientResourceIT.java
```diff
@@ -37,9 +38,9 @@ public void shouldUpdateJaegerAsTracer() {
                 .then()
                 .statusCode(HttpStatus.SC_OK)
                 .body(equalTo("I'm a client"));
-
         await().atMost(30, TimeUnit.SECONDS).untilAsserted(() -> {
-            get(jaeger.getTraceUrl() + "?service=" + SERVICE_NAME)
+            Response response = get(jaeger.getTraceUrl() + "?service=" + SERVICE_NAME);
```
pjgg, line 42: ...and what about this approach?
```
 String traceURL = jaeger.getTraceUrl() + "?service=" + SERVICE_NAME;
            get(traceURL)...
```

quarkus-test-core/src/main/java/io/quarkus/test/services/URILike.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public int getPort() {
+        return wrapped.getPort();
+    }
+
+    /**
+     * @return "URI" formatted in style, preferred by RestAssured library(e.g. "http://localhost")
+     */
+    public String getRestAssuredStyleUri() {
```
pjgg, line 100: What do you think about this other method name?

`getRestAssuredStyleUri` -> `toRestAssuredUri`

quarkus-test-core/src/test/java/io/quarkus/test/services/URILikeTest.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.test.services;
+
+import static org.junit.jupiter.api.Assertions.*;
```
pjgg, line 3: is not used !

quarkus-test-core/src/test/java/io/quarkus/test/services/URILikeTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class URILikeTest {
+
+    @Test
+    void string() {
```
pjgg, line 11: Better test name, something like `stringSerialization` ...

pjgg: Could we verify some border conditions, for example, If I create a `URILike` without scheme...make sense? and IF does which should be the expected result? 

`URILike uri = new URILike("", "localhost", 1102, null);` -> `://localhost:1102` ?

or the port has some limits, could we have a negative port? ... think in border conditions and be sure that we have the expected response for these cases.

quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/OpenShiftQuarkusApplicationManagedResource.java
```diff
@@ -74,20 +75,19 @@ public void stop() {
...
-    public String getHost(Protocol protocol) {
-        validateProtocol(protocol);
-        return untilIsNotNull(() -> client.url(model.getContext().getOwner()),
-                AwaitilitySettings.defaults().withService(getContext().getOwner()));
-    }
-
-    @Override
-    public int getPort(Protocol protocol) {
+    public URILike getURI(Protocol protocol) {
+        if (protocol == Protocol.HTTPS && !client.isServerlessService(model.getContext().getName())) {
```
pjgg, line 79: I think that this validation could be in a private method as it was in main branch, WDYT?

pjgg (changes requested):
Overall looks good but we need also to run OCP coverage

quarkus-test-core/src/main/java/io/quarkus/test/services/URILike.java
fedinskiy: "Rest Assured URI" looks like it is URI, which points to Rest Assured while this method returns URI, which can be accepted by RestAssured. Probably we can use getUriForRestAssured, but in one or two places it used for other things, so the name would be a bit misleading

examples/jaeger/src/test/java/io/quarkus/qe/ClientResourceIT.java
fedinskiy: Removed this stuff altogether

quarkus-test-core/src/test/java/io/quarkus/test/services/URILikeTest.java
fedinskiy: Renamed and added some tests. Thank you to pointing me to this

fedinskiy:
run tests

quarkus-test-core/src/test/java/io/quarkus/test/services/URILikeTest.java
```diff
@@ -0,0 +1,82 @@
+package io.quarkus.test.services;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class URILikeTest {
```
pjgg, line 9: Too much better!

pjgg (approved):
LGTM

PR #558, opened 2022-09-11
Fix OC failures - migrate io.fabric8 Kubernetes Client to 6.1.1

michalvavrik:
### Summary

`DefaultOpenShiftClient` has been [deprecated](https://developers.redhat.com/articles/2022/07/15/new-http-clients-java-generator-and-more-fabric8-600#other_improvements) in 6.1.1 and calling `adapt` is throwing `NoSuchMethodError` exception as you can see below:
```
java.lang.NoSuchMethodError: io.fabric8.openshift.client.NamespacedOpenShiftClient.adapt(Ljava/lang/Class;)Ljava/lang/Object;
```
Currently the Test Framework is using Quarkus 2.12.0.Final that provides `io.fabric8.kubernetes-client` in version `5.12.3`, however when running our TS with Quarkus 999-SNAPSHOT, `6.1.1` is used. We need to make our TFW compatible with 6.1.1, therefore I declare the version explicitly. It's necessary to use BOM as transitive dependencies are otherwise managed by the Quarkus 2.12.0.Final. In order to override Quarkus dependency management, I had to place the BOM to the `quarkus-test-kubernetes` and `quarkus-test-openshift`.

- [x] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,9 @@
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
         <quarkus.platform.version>2.12.0.Final</quarkus.platform.version>
+        <!--  TODO: once we migrate to Quarkus 2.13, delete 'kubernetes-client-bom' and 'dekorate' from dependency management (see below)  -->
+        <kubernetes-client.version>6.1.1</kubernetes-client.version>
+        <dekorate.version>3.0.0</dekorate.version>
         <exclude.tests.with.tags>quarkus-cli</exclude.tests.with.tags>
         <include.tests>**/*IT.java</include.tests>
         <exclude.openshift.tests>**/OpenShift*IT.java</exclude.openshift.tests>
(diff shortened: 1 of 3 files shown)


michalvavrik:
run tests

pom.xml
```diff
@@ -49,6 +49,10 @@
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
         <quarkus.platform.version>2.12.0.Final</quarkus.platform.version>
+        <!--  'kubernetes-client' should be managed by Quarkus once we migrate to 2.13  -->
```
jsmrcka, line 52: I would add a `TODO`, in order to make it more visible for whoever will perform the upgrade to Quarkus 2.13.

(The same would go for the descendant POMs, but the current changes in them can be discarded - see the comment in `quarkus-test-kubernetes/pom.xml`.)

quarkus-test-kubernetes/pom.xml
```diff
@@ -8,6 +8,18 @@
     </parent>
     <artifactId>quarkus-test-kubernetes</artifactId>
     <name>Quarkus - Test Framework - Kubernetes</name>
+    <dependencyManagement>
```
jsmrcka, line 11: Currently, OpenShift CI is failing because other modules which depend on `quarkus-test-openshift` and/or `quarkus-test-kubernetes` (e.g. `examples/greetings`) still use the old version defined by Quarkus dependency management.

In order to make our override global for the whole TF, I suggest placing this import in **parent POM** and making sure that it is listed **before** the import of `quarkus-bom` (that effectively means making it the first import in the `dependencyManagement` section).

That way, no further imports in descendant POMs should be needed, so this one and the other one in `quarkus-test-openshift` may be removed.

jsmrcka (changes requested):
Thanks a lot for this PR! I've been trying to unblock myself from the k8s upgrade, but you were quicker. :+1: :relieved:

quarkus-test-kubernetes/src/main/java/io/quarkus/test/bootstrap/inject/KubectlClient.java
```diff
@@ -67,17 +67,22 @@ public final class KubectlClient {
     private static final int HTTP_PORT_DEFAULT = 80;
 
     private final String currentNamespace;
-    private final DefaultOpenShiftClient masterClient;
-    private final NamespacedOpenShiftClient client;
+    private final KubernetesClient masterClient;
+    private final io.fabric8.openshift.client.OpenShiftClient client;
```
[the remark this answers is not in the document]
michalvavrik: Sure, thanks

pom.xml
michalvavrik: Yes, thank you

quarkus-test-kubernetes/pom.xml
michalvavrik: That won't do (among others, it causes compilation error) as `quarkus-kubernetes-client-deployment` will only work with 6.1.1 after some changes (see `deps: Bump kubernetes-client-bom from 5.12.2 to 6.1.1`), thus it's vital we isolate migration to `quarkus-test-openshift` and `quarkus-test-kubernetes`. I am sorry I didn't mention the errors you are referring to before as I only tested this fix with the test suite and there was no problem as it uses 999-SNAPSHOT.

We can't force version to just `openshift-client` and `knative-client` as it will always conflict with 5.12.3 brought in by the Quarkus. I'll think about it tomorrow, only thing that worked for me this far is to bring manually changes to `KubernetesClientProcessor` and `dekorate`, but it's very ugly.

michalvavrik: @jsmrcka bringing in dekorate deps together with your solution fixed compilation

michalvavrik:
run tests

michalvavrik:
10 OC failures goes down to usage of Quarkus 2.12. It's way less then last time though.

quarkus-test-kubernetes/pom.xml
jsmrcka: You're right, I didn't properly test the solution suggested by me. It ends up throwing another `NoSuchMethodError`  e.g. when running `mvn clean install -Pframework,examples`.

jsmrcka (approved):

jsmrcka:
Resolving OCP issues would be to costly, let's merge and wait for 2.13.

PR #571, opened 2022-09-19
Improve helm usability by adding new commands and shortcuts

pjgg:
### Summary

- HelmClient `getChartsNames` method was created as a Shortcut method in order to retrieve current chartsNames
- HelmClient `chartDependencyUpdate` and `chartDependencyBuild` was added as new supported helmClient methods.
- Re-Enable `OpenShiftQuarkusHelmClientIT` scenario
- `quarkus-helm"` Tag was used at the beginning in order to enable/disable these tests if a specific `profile/tags` was enabled (currently, we are doing something like this in Quarkus-cli scenario). However, Quarkus-helm is not ignored anymore and always are going to be triggered, so this tag is not required anymore 

- [X] Refactoring

### Checklist:
- [X] Example scenarios has been updated / added

### DIFF
--- a/examples/quarkus-helm/src/main/resources/application.properties
+++ b/examples/quarkus-helm/src/main/resources/application.properties
@@ -3,13 +3,10 @@
 quarkus.openshift.route.expose=true
 quarkus.kubernetes.ingress.expose=true
 
-# TODO https://github.com/quarkiverse/quarkus-helm/issues/41
-#quarkus.container-image.image=quay.io/quarkusqeteam/examples-quarkus-helm:v1
 quarkus.container-image.registry=quay.io
 quarkus.container-image.group=quarkusqeteam
 quarkus.container-image.name=examples-quarkus-helm
 quarkus.container-image.tag=v1
-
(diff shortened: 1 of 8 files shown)


examples/quarkus-helm/src/main/resources/application.properties
```diff
@@ -3,13 +3,10 @@
 quarkus.openshift.route.expose=true
 quarkus.kubernetes.ingress.expose=true
 
-# TODO https://github.com/quarkiverse/quarkus-helm/issues/41
-#quarkus.container-image.image=quay.io/quarkusqeteam/examples-quarkus-helm:v1
```
fedinskiy, line 7: I see, that the issue is fixed, but we are not using this image. Is this expected?

pjgg: Yes, that is a good point, I didn't want to use this way to set the docker image path/name because this way is used in [Quarkus-test-suite helm module](https://github.com/quarkus-qe/quarkus-test-suite/pull/835/files#diff-4337e56031e9dcce3124cb9f641b4ef17906bc5db5c4158f4127397803401e82R3), so by this way we are covering both scenarios :)

fedinskiy (approved):

PR #581, opened 2022-09-28
Make OpenShift service expose configurable

michalvavrik:
### Summary

Now it's not possible to deploy app to OpenShift using existing deployment strategies without exposing service, however sometimes you need more sophisticated route to expose/or you don't want to expose service at all. In my case, I'm deploying Quarkus serverless function that exist within Knative eventing namespace and I'm need to communicate only with broker, not this service (so far, I didn't even found a way to expose it and I don't think it's desirable). This PR adds flag that allows you to disable exposing service like this:

`ts.app.openshift.expose-service=false`

If the property is not set, nothing changed at all. I also merged two identical methods into one (refactoring part).

- [ ] Dependency update
- [x] Refactoring

### Checklist:


michalvavrik:
I have to think about `waitUntilServiceIsStarted` too.

PR #601, opened 2022-10-21
Remove Quarkus-helm from example modules

pjgg:
There is an issue on Upstream related to dekorate that make this module fails on compile time
Ref: quarkiverse/quarkus-helm#114


michalvavrik (approved):
lgtm

PR #719, opened 2023-03-26
Revert "Enable database native tests"

michalvavrik:
Reverts quarkus-qe/quarkus-test-framework#714

We are building native executable and then running DEV mode test. The tests we are running has absolutely nothing to do with a native mode, the PR doesn't actually run database tests in native mode as it claims (**doesn't use native executable**) and we should revert it as it waste running time and resources.


michalvavrik:
hey @gtroitsk , just fyi - that's not a criticism, this one is easy to miss :-)

pjgg:
Could you add a comment (or a disable reason) that explains why these tests are disabled in order to avoid this missunderstood in the future? @michalvavrik

michalvavrik:
Sure, will do @pjgg .

PR #764, opened 2023-04-26
Fix/mysql fips

fedinskiy:
### Summary

 Add FIPS-specific query parameters
    
  Fixes https://github.com/quarkus-qe/quarkus-test-suite/issues/756

Also, fix for openshift-template typo

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/URILike.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/URILike.java
@@ -71,6 +71,18 @@ public class URILike {
         return new URILike(this.scheme, withPath);
     }
 
+    public URILike withQuery(String query) {
+        URI wrapped = this.wrapped;
+        var withPath = createURI(wrapped.getScheme(),
+                wrapped.getUserInfo(),
+                wrapped.getHost(),
+                wrapped.getPort(),
+                wrapped.getPath(),
+                query,
+                wrapped.getFragment());
+        return new URILike(this.scheme, withPath);
+    }
+
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder();
--- a/quarkus-test-service-database/src/main/java/io/quarkus/test/bootstrap/MySqlService.java
+++ b/quarkus-test-service-database/src/main/java/io/quarkus/test/bootstrap/MySqlService.java
@@ -7,6 +7,7 @@ public class MySqlService extends DatabaseService<MySqlService> {
     static final String PASSWORD_ROOT_PROPERTY = "MYSQL_ROOT_PASSWORD";
     static final String DATABASE_PROPERTY = "MYSQL_DATABASE";
     static final String JDBC_NAME = "mysql";
+    private static final String FIPS_OPTIONS = "useSSL=false&allowPublicKeyRetrieval=True";
 
     @Override
     protected String getJdbcName() {
@@ -21,5 +22,21 @@ public class MySqlService extends DatabaseService<MySqlService> {
         withProperty(DATABASE_PROPERTY, getDatabase());
 
         return super.onPreStart(action);
+    }
+
+    public String getJdbcUrl() {
+        return getURI()
+                .withScheme("jdbc:" + getJdbcName())
+                .withPath("/" + getDatabase())
+                .withQuery(FIPS_OPTIONS)
+                .toString();
+    }
+
+    public String getReactiveUrl() {
+        return getURI()
+                .withScheme(getJdbcName())
+                .withPath("/" + getDatabase())
+                .withQuery(FIPS_OPTIONS)
+                .toString();
     }
 }
--- a/quarkus-test-openshift/src/main/resources/quarkus-registry-openshift-template.yml
+++ b/quarkus-test-openshift/src/main/resources/quarkus-registry-openshift-template.yml
@@ -55,7 +55,6 @@ items:
             - containerPort: ${INTERNAL_PORT}
               name: "http"
               protocol: "TCP"
-              ports:
             - containerPort: ${MANAGEMENT_PORT}
               name: "management"
               protocol: "TCP"


quarkus-test-service-database/src/main/java/io/quarkus/test/bootstrap/MySqlService.java
```diff
@@ -22,4 +23,20 @@ public MySqlService onPreStart(Action action) {
 
         return super.onPreStart(action);
     }
+
+    public String getJdbcUrl() {
+        return getURI()
+                .withScheme("jdbc:" + getJdbcName())
+                .withPath("/" + getDatabase())
+                .withQuery(FIPS_OPTIONS)
+                .toString();
```
michalvavrik, line 32: there is no FIPS in `getJdbcUrl` name, can we only add this options if environment variable `FIPS` is set to `fips`? I don't want to alter connection string unless we really need to

michalvavrik (changes requested):

quarkus-test-service-database/src/main/java/io/quarkus/test/bootstrap/MySqlService.java
fedinskiy: This behavior should work in ordinary mode too.
And reading environment variables in a method is sort of antipattern, since it is not explicit

quarkus-test-service-database/src/main/java/io/quarkus/test/bootstrap/MySqlService.java
```diff
@@ -22,4 +23,20 @@ public MySqlService onPreStart(Action action) {
...
+    }
+
+    public String getReactiveUrl() {
+        return getURI()
+                .withScheme(getJdbcName())
```
michalvavrik, line 37: at which point is added `vertx-reactive:` before `mysql`?

michalvavrik: okay,  I can see `io.quarkus.test.bootstrap.DatabaseService#getReactiveUrl` does same, so I suppose it do work

quarkus-test-core/src/main/java/io/quarkus/test/services/URILike.java
```diff
@@ -71,6 +71,18 @@ public URILike withPort(int port) {
         return new URILike(this.scheme, withPath);
     }
 
+    public URILike withQuery(String query) {
```
michalvavrik, line 74: please add new method (getter) `getQuery` for cases someone want to join query with the previous one

fedinskiy: Do we have such cases?

fedinskiy: I would prefer to keep api size as small as possible.
For context, existing `getUserInfo` isn't used anywhere outside unit tests for this class.

michalvavrik: ok

michalvavrik (approved):
I dislike always adding FIPS_OPTIONS, but approved, thanks for the fix

quarkus-test-service-database/src/main/java/io/quarkus/test/bootstrap/MySqlService.java
```diff
@@ -7,6 +7,7 @@ public class MySqlService extends DatabaseService<MySqlService> {
     static final String PASSWORD_ROOT_PROPERTY = "MYSQL_ROOT_PASSWORD";
     static final String DATABASE_PROPERTY = "MYSQL_DATABASE";
     static final String JDBC_NAME = "mysql";
+    private static final String FIPS_OPTIONS = "useSSL=false&allowPublicKeyRetrieval=True";
```
michalvavrik, line 10: can you add TODO to switch from `useSSL` to `sslMode` when all our MySQL versions are greater or equal to 8.0.13?

rsvoboda:
Going forward I think we should consider different approach for FIPS enabled environment. We may have other environments with some specifics and we should have a more general way for treating different environments.

I have no concrete solution in mind, maybe something like environment based enrichers. These enricher can be applied in the services, or Enrichers would be just information holders and each service could adapt on their own.
We need to decide if we are running against bare metal or OCP, we could also track target environment specifics. We may need to run against FIPS enabled OCP which could some specifics comparing to fips enabled RHEL.

@mjurc @michalvavrik wdyt?

About this PR: I don't like that the jdbc url is always altered even though fips enable use-case is not the major one. We must just hope it won't introduce side effects in the future.

michalvavrik:
> I have no concrete solution in mind, maybe something like environment based enrichers. These enricher can be applied in the services, or Enrichers would be just information holders and each service could adapt on their own. We need to decide if we are running against bare metal or OCP, we could also track target environment specifics. We may need to run against FIPS enabled OCP which could some specifics comparing to fips enabled RHEL.

IMO it's only worth effort when we have more FIPS specific customizations because it will come with a cost - boilerplate/extra code for abstraction

as far as this PR is concerned @fedinskiy, approval still holds, but I realized 2 things:

- I'd like to run OpenShift tests because of template change (I didn't write `run tests` as I hope you will address my comment)
- if I read this https://github.com/quarkusio/quarkus/issues/32910#issuecomment-1523711393 correctly, your fix won't help for the moment being as it won't work anyway (not in JVM mode, not in native mode). Am I wrong? Shouldn't we wait and test your changes when upstream issue is fixed so that we have way to verify them (don't get me wrong, I trust you)

michalvavrik:
ouch, so just mentioning r*u*n tests anywhere in code triggers job :-D

rsvoboda:
@fedinskiy / @michalvavrik is the fail related to the change / can be the PR merged?

michalvavrik:
> @rsvoboda the fail is not related, but I think the PR **should not** be merged (see above).

+1 please don't merge; it won't work in native or JVM anyway till upstream provide supported (customer friendly) way, I vote for waiting to upstream issue fix.

rsvoboda:
> I vote for waiting to upstream issue fix

So we should close this PR for now?

michalvavrik:
> > I vote for waiting to upstream issue fix
> 
> So we should close this PR for now?

That's what I tried not to say(but thought), maybe @fedinskiy have other plans as we watch upstream issue closely

fedinskiy:
This PR tries to solve https://github.com/quarkusio/quarkus/issues/32910 in a wrong way, so waiting for proper upstream fix is probably a better idea.

PR #781, opened 2023-05-12
Do not manage HTMLUnit and Jaeger client dependencies

michalvavrik:
### Summary

We don't use Jaeger client and HTMLUnit in Test Framework and shouldn't waste dependabot runs on them. If anything that depends on our framework actually needs this dependencies, it should manage them itself.

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,12 +40,10 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <surefire-plugin.version>3.1.0</surefire-plugin.version>
         <jacoco-maven-plugin.version>0.8.10</jacoco-maven-plugin.version>
-        <jaeger.version>1.8.1</jaeger.version>
         <prometheus.simpleclient_pushgateway.version>0.16.0</prometheus.simpleclient_pushgateway.version>
         <maven-checkstyle-plugin.version>3.2.1</maven-checkstyle-plugin.version>
         <formatter-maven-plugin.version>2.22.0</formatter-maven-plugin.version>
         <impsort-maven-plugin.version>1.9.0</impsort-maven-plugin.version>
-        <htmlunit.version>2.70.0</htmlunit.version>
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
         <quarkus.platform.version>3.0.3.Final</quarkus.platform.version>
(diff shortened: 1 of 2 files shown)


michalvavrik:
I'll rebase to resolve conflicts.

pom.xml
```diff
@@ -172,23 +170,6 @@
...
-                        <groupId>org.apache.tomcat.embed</groupId>
-                        <artifactId>tomcat-embed-core</artifactId>
-                    </exclusion>
-                </exclusions>
-            </dependency>
-            <dependency>
-                <groupId>net.sourceforge.htmlunit</groupId>
-                <artifactId>htmlunit</artifactId>
-                <version>${htmlunit.version}</version>
-            </dependency>
```
jsmrcka, line 191: This dependency is used in QE TS in several modules, e.g. https://github.com/quarkus-qe/quarkus-test-suite/blob/4ad7c428a48f4dbaaae90cbd88066d9c12e1225a/security/keycloak-jwt/pom.xml#L22.

Unless we manage the dependency there (likely in the TS's root pom), it would not compile.

Also, there is a logging level set in TF: https://github.com/quarkus-qe/quarkus-test-framework/blob/7a6f350295e4633dd10d6650aec8c0181d0c5cd2/quarkus-test-core/src/main/resources/logging.properties#L4.
If the dependency is removed from TF, we should IMO remove this property as well.
But that also means figuring out a way to apply it in the TS. Otherwise, the `htmlunit` logging output would start littering TS logs.

jsmrcka (changes requested):

pom.xml
michalvavrik: @jsmrcka I can drop logging, although I don't see anything bad on setting it here because this way, it is set for all tests that could be using it (though TBH, I think it's remnant as I got rid of HTML unit from test-core and now it is not used there!). 

But whether or not TS decides to manage something, it's not FW business. It's like if we decided to manage here all dependencies that could be used by TS, how does it make sense to run CI for something we don't use here? We need to raise dependency version in the place where it is used because this way, we will recognize problem with new HTML Unit artifact version on tests.

michalvavrik: I'll have a look how HTML Unit is used in TS and come back to you.

michalvavrik: before I do - my opinion is that setting is here is not a big deal. The relation between managing dependency here and setting it here is made by you and me. There is none IMHO because it is different to if we were actually using dependency here.

michalvavrik: Okay, I had a look. We only need to repeat property in TS 6 times, I suggest to drop it.

michalvavrik: Done.

jsmrcka: > But whether or not TS decides to manage something, it's not FW business.

I'm not arguing about that.
My point is that we should prepare a PR which adds the management of this dependency in TS and do it rather now than later.
That way we don't forget about it and avoid TS failures because of the missing dependency.

I agree with dropping the property from FW.
It would be nice to be able to set it just once in TS, but I haven't figured out a way how to do it. But 6 times is IMO fine too.

michalvavrik: Sure, here you go https://github.com/quarkus-qe/quarkus-test-suite/pull/1226

jsmrcka: Thanks. :+1:

PR #824, opened 2023-07-03
Improve detection of Native mode

fedinskiy:
Previously, @DisabledOnNative annotations used package type to detect whether the test is executed in Native mode. Unfortunately, this doesn't work for modules, which have their native mode disabled (e.g. quarkus-cli[1]). This changes uses Maven profiles to detect Native mode, the same way it worked[1] for quarkus-cli

[1] https://github.com/Sgitario/quarkus-test-suite/blob/afebdd2e187c2c3edcc53fc6ce2cff4b8c629a17/quarkus-cli/pom.xml#L20-L29
[2] https://github.com/quarkus-qe/quarkus-test-suite/pull/1230

### Summary

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/DisabledOnNativeCondition.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/DisabledOnNativeCondition.java
@@ -1,5 +1,10 @@
 package io.quarkus.test.scenarios.annotations;
 
+import static org.junit.jupiter.api.extension.ConditionEvaluationResult.disabled;
+import static org.junit.jupiter.api.extension.ConditionEvaluationResult.enabled;
+
+import java.util.Objects;
+
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
@@ -10,10 +15,13 @@ public class DisabledOnNativeCondition implements ExecutionCondition {
 
     @Override
     public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
-        if (!QuarkusProperties.isNativePackageType()) {
-            return ConditionEvaluationResult.enabled("It's not running the test on Native");
+        if (QuarkusProperties.isNativePackageType()) {
+            return disabled("It's running the test on Native, detected through package type");
         }
-
-        return ConditionEvaluationResult.disabled("It's running the test on Native");
+        String property = System.getProperty("profile.id");
+        if (Objects.equals(property, "native")) {
+            return disabled("It's running the test on Native, detected through build profile");
+        }
+        return enabled("It's not running the test on Native");
     }
 }


michalvavrik (commented):
@fedinskiy change is fine (and I will approve you if you insist), but if we _disable_ native build by setting package type to `fast-jar`, we are still running Quarkus build. What for, if we know on level of Maven that we don't want to run tests there? I'd prefer to replace all occurrences of `<quarkus.package.type>fast-jar</quarkus.package.type>` with `<quarkus.build.skip>true</quarkus.build.skip>` instead? This way, we can still keep using package type and we won't experience issue you describe. 

Please look at https://quarkus.io/guides/building-native-image that shows that native executable is produced when `./gradlew build -Dquarkus.package.type=native`, so if we stick with current solution, it will also work for Gradle.

fedinskiy:
@michalvavrik  I gave this a thought and find two arguments for a change:
1. Current behavior is surprising, and an API should not be surprising. 
2. There may be a future module, which would need current behavior changed to work properly. It should have three properties: a) has native build disabled via `fast-jar` trick b) requires running tests when native profile enabled c)  requires disabling some tests when native profile disabled. 
I would try to imagine such module. If the description is realistic, when we need to implement these changes, if it is too far-fetched, when we can safely discard them.
Suppose, that we have module, which runs tests over its own `src/main` code and over some external applications (eg via `@GitRepositoryQuarkusApplication`). There are some problems/compilation speed issues with the module, so we add `fast-jar` property to not compile `src/main` into native and disable tests, which check `src/main`. But we still need to run tests over external applications and pass the `native` parameter to them, so we do not want to have `quarkus.build.skip` during native compilation.
Is such module possible? I would say yes, and it is similar to existing `narayna-transactions`, but the decision is up to you.

michalvavrik (approved):
I don't really agree with arguments, but IMO there is not one right answer.  LGTM, let's not spend time on this.

Few thoughts that can be ignored:

We are going to ignore possible Gradle scenarios? You will need to pass `-Dnative` even though you don't want to active Maven profile. Package type felt more natural to me.

> Current behavior is surprising, and an API should not be surprising.

Well, `@DisabledOnNative` should refer to native tests (with native executable) and that is not same as native profile. When we check package type, it actually is concerned on whether it is native test (uses native executable).

> But we still need to run tests over external applications and pass the native parameter to them, so we do not want to have quarkus.build.skip during native compilation.

`quarkus.build.skip` skips Quarkus build, not tests, so external application would still receive `-Dnative`. Also FYI when Quarkus build is skipped and executable is not found, our FW should simply build it itself as it does when custom build is needed by test class (e.g. when build time property is set).

btw. upstream uses `io.quarkus.deployment.pkg.steps.NativeBuild` to run certain steps for native build only and they also use package config for it (not profile).

fedinskiy:
Just in case: package type doesn't go anywhere, profile check is after the package type check.
We do not have gradle modules in the ts or fw . External Gradle scenarios would require some magic anyway.

> `quarkus.build.skip` skips Quarkus build, not tests, so external application would still receive `-Dnative`

In that case, I see no reason in this change. You persuaded me :)

PR #831, opened 2023-07-12
Update Quarkus to version 3.2.0.Final

fedinskiy:
### Summary
The title says it all

- [x] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/examples/quarkus-helm/pom.xml
+++ b/examples/quarkus-helm/pom.xml
@@ -5,12 +5,12 @@
         <groupId>io.quarkus.qe</groupId>
         <artifactId>quarkus-examples-parent</artifactId>
         <version>1.3.1.Beta21-SNAPSHOT</version>
-        <relativePath>../</relativePath>
+        <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>examples-quarkus-helm</artifactId>
     <name>Quarkus - Test Framework - Examples - Quarkus HELM</name>
     <properties>
-        <quarkus-helm.version>1.0.8</quarkus-helm.version>
+        <quarkus-helm.version>1.0.9</quarkus-helm.version>
     </properties>
 
     <dependencies>
--- a/examples/consul/src/test/java/io/quarkus/qe/OpenShiftUsingExtensionBuildTimeGreetingResourceIT.java
+++ b/examples/consul/src/test/java/io/quarkus/qe/OpenShiftUsingExtensionBuildTimeGreetingResourceIT.java
@@ -2,7 +2,9 @@ package io.quarkus.qe;
 
 import io.quarkus.test.scenarios.OpenShiftDeploymentStrategy;
 import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.test.scenarios.annotations.DisabledOnQuarkusVersion;
 
 @OpenShiftScenario(deployment = OpenShiftDeploymentStrategy.UsingOpenShiftExtension)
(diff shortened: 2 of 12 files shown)


examples/quarkus-helm/pom.xml
```diff
@@ -5,12 +5,12 @@
         <groupId>io.quarkus.qe</groupId>
         <artifactId>quarkus-examples-parent</artifactId>
         <version>1.3.1.Beta21-SNAPSHOT</version>
-        <relativePath>../</relativePath>
+        <relativePath>../pom.xml</relativePath>
```
michalvavrik, line 8: I have tried it without this change (pom.xml) and it worked. Other modules are using `../`, could you provide context on why it should be different in this particular module?

.github/workflows/pr.yaml
```diff
@@ -17,6 +17,9 @@ jobs:
           java-version: ${{ matrix.java }}
           check-latest: true
           cache: 'maven'
+      - name: Pre-build framework
+        run: |
+          mvn -V -B -s .github/mvn-settings.xml install -Pframework -DskipTests -DskipITs
```
michalvavrik, line 22: I would expect this for Maven to handle dependencies between module and determine order. This step is fine (I'm not rejecting changes!), but I'll definitely try to look at this later because it seems weird. Or am I missing something?

examples/quarkus-helm/pom.xml
fedinskiy: The thing is, I can not reproduce the failure at all, with or without any changes except 3.1.0->3.2.0. If the checks would be green, I will start dropping rest of the changes. I suppose, update of help should be enough, but I am not 100% sure.

.github/workflows/pr.yaml
fedinskiy: You don't this is a weird step. I will try to drop it, if other changes would be enough.

examples/quarkus-helm/pom.xml
michalvavrik: Sure, makes sense

.github/workflows/pr.yaml
michalvavrik: Ok, thx

fedinskiy: It looks like this was not required after all

examples/quarkus-helm/pom.xml
fedinskiy: Helm update is not enough, validation requires this change too.

I do not know, why it needs `../pom.xml` and not just `../`, but at least Idea complains about it, so it looks like `../` is not guaranteed to work at all and now it finally stopped :)

fedinskiy: And maven doc[1] explicitly says: "The relative path of the parent <code>pom.xml</code> file within the check out."

[1] https://maven.apache.org/xsd/maven-4.0.0.xsd

fedinskiy:
run tests

examples/quarkus-helm/pom.xml
michalvavrik: Oki

michalvavrik:
@fedinskiy I glanced at OC results, looks related, WDYT? This is not a good release for OpenShift :-D

fedinskiy:
@michalvavrik this is quite peculiar, since all failures has `Connection was closed` in common.

My bet is on https://github.com/quarkusio/quarkus/issues/34276 which should be fixed in 3.2.1.Final

fedinskiy:
run tests

michalvavrik (approved):

PR #864, opened 2023-08-25
Disable QuarkusScenario tests automatically when Linux containers are not available but the test requires them

michalvavrik:
### Summary

Tests are automatically disabled when Linux containers are not available,  but the test has managing resource that requires Linux containers. Detecting when Quarkus CLI starts in DEV mode will require additional effort.

- [ ] Dependency update
- [x] Refactoring
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/scenarios/execution/condition/AbstractQuarkusScenarioContainerExecutionCondition.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/scenarios/execution/condition/AbstractQuarkusScenarioContainerExecutionCondition.java
@@ -1,0 +1,121 @@
+package io.quarkus.test.scenarios.execution.condition;
+
+import static java.lang.Boolean.FALSE;
+import static java.lang.Boolean.TRUE;
+import static java.lang.String.format;
+import static java.lang.String.join;
+import static org.junit.jupiter.api.extension.ConditionEvaluationResult.disabled;
+import static org.junit.jupiter.api.extension.ConditionEvaluationResult.enabled;
+
+import java.lang.reflect.AnnotatedElement;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import org.jboss.logging.Logger;
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.utils.Command;
+
+public abstract class AbstractQuarkusScenarioContainerExecutionCondition implements QuarkusScenarioExecutionCondition {
+
+    private static final Logger LOG = Logger.getLogger(AbstractQuarkusScenarioContainerExecutionCondition.class.getName());
+    private static final ConditionEvaluationResult CONDITION_NOT_MATCHED = enabled("This condition should "
+            + "only be applied on test classes annotated with the '@QuarkusScenario' annotation");
+    private static final ConditionEvaluationResult ENV_SUPPORTS_LINUX_CONTAINERS = enabled("Environment supports"
+            + " Linux containers");
+    private static final String ENV_DOES_NOT_SUPPORT_LINUX_CONTAINERS = "Test class '%s' requires Linux containers, "
+            + "but the environment does not support them";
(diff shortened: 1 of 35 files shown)


quarkus-test-containers/src/main/java/io/quarkus/test/services/execution/conditions/EnableOnLinuxContainersWhenContainersRequiredCondition.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private static boolean isQuarkusScenario(AnnotatedElement annotatedElement) {
+        return annotatedElement.isAnnotationPresent(QuarkusScenario.class);
+    }
+
+    private static Collection<String> runDockerOsTypeInfoCommand(Class<?> testClass) {
+        try {
+            var output = new ArrayList<String>();
+            new Command("docker", "info", "--format", "'{{.OSType}}'")
```
mjurc, line 102: I'd presume that podman returns the same response, correct? Just to make sure.

Maybe it would be nice to run the test suite on podman too.

mjurc (commented):
The change makes sense and the conditions are correct, as far as I can see from the review.

Can we have some runs with test suite on baremetal with docker/podman, too? Perhaps it would be for the better to try this on OCP jobs also.

michalvavrik:
run tests

quarkus-test-core/src/main/java/io/quarkus/test/scenarios/execution/condition/AbstractQuarkusScenarioContainerExecutionCondition.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private static boolean isQuarkusScenario(AnnotatedElement annotatedElement) {
+        return annotatedElement.isAnnotationPresent(QuarkusScenario.class);
+    }
+
+    private static Collection<String> runDockerOsTypeInfoCommand() {
+        try {
+            var output = new ArrayList<String>();
+            new Command("docker", "info", "--format", "'{{.OSType}}'")
```
mjurc, line 90: I think `{{.OSType}}` is not present on podman - I tried `podman info` locally without any such line being provided. 

I think the podman failures in CI are caused by that.

mjurc (changes requested):
Actually I think my last review comment will require changes.

quarkus-test-core/src/main/java/io/quarkus/test/scenarios/execution/condition/AbstractQuarkusScenarioContainerExecutionCondition.java
michalvavrik: > I think {{.OSType}} is not present on podman - I tried podman info locally without any such line being provided.

I see, I based my input on stuff from Podman on Windows, either it differs or I misunderstood the info.

> I think the podman failures in CI are caused by that.

sure

> Mind you - https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-core/src/main/java/io/quarkus/test/utils/Command.java#L72 throws runtime exception in case the return value of launched command is non-zero, which I think could be an issue for this as well - as runtime exceptions are notoriously hard to catch. This will result in exception rather than skipped test.

Not sure why we couldn't catch `RuntimeException`, I'm probably missing something, but anyway, exception handling is expensive and I would rather avoid it. I checked how Quarkus implements it `io.quarkus.runtime.util.ContainerRuntimeUtil#detectContainerRuntime(boolean)` and it doesn't feel nor efficient or suitable for our case. I suggest to check `DOCKER_HOST` which should do on Linux (and is quicker then executing command) then check docker version for `podman` keyword then do the same info command. It's one more command then I'd like, but I couldn't figure how to use `missingkey` in Go templates nor how to convince JSON format to don't complain about missing key.

I tested it on Podman instance manually, but will re-trigger Podman job as well. Thanks for fixing it.

michalvavrik:
run tests

mjurc:
run tests

mjurc (approved):
Thanks. Think this looks fine now, although that `catch (Exception ex)` is kinda nasty.

Let's get the CI run green - it seems that the test clusters had full objectstore after last week.

michalvavrik:
All green, merging, thanks for the review.

PR #885, opened 2023-09-21
Revert "Bump io.confluent:kafka-avro-serializer from 7.3.4 to 7.5.0"

michalvavrik:
Reverts quarkus-qe/quarkus-test-framework#862

I made mistake, see https://github.com/quarkus-qe/quarkus-test-framework/pull/862#issuecomment-1729318761.

### DIFF
--- a/examples/kafka-registry/pom.xml
+++ b/examples/kafka-registry/pom.xml
@@ -10,7 +10,7 @@
     <artifactId>examples-kafka-registry</artifactId>
     <name>Quarkus - Test Framework - Examples - Kafka with Registry</name>
     <properties>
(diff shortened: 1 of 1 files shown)


michalvavrik:
> I'm okay with this - the way I understand this is this causes native checks to fail.

yep
 
> So, next steps are about enhancing our PR CI with native checks?

I think so. The problem is that I know about the missing native step and keep forgetting to run it locally when reviewing PRs. So I think we need that.

michalvavrik:
I've opened issue https://github.com/quarkus-qe/quarkus-test-framework/issues/886.

PR #893, opened 2023-10-02
Enable to use java preview while executing built jar file.

jedla97:
### Summary

Hi, this PR adding option run built jar with preview argument. This can be useful when someone trying the JEPs which are marked as preview. What I tried when using framework with quarkus dev mode it pass the property `-D--enable-preview` but when using standard mode the property is not picked so jar failing to execute.

Running `mvn -B -fae clean install -Pframework,examples -Dvalidate-format -Dquarkus.platform.version="999-SNAPSHOT" -pl examples/database-mysql -D--enable-preview `

Before this PR result in produced command `java -D<some Quarkus properties> -jar quarkus-test-framework/examples/database-mysql/target/quarkus-app/quarkus-run.jar`

After this PR result in produced command `java --enable-preview -D<some Quarkus properties> -jar quarkus-test-framework/examples/database-mysql/target/quarkus-app/quarkus-run.jar`

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
@@ -7,10 +7,14 @@ import java.util.List;
 
 import org.apache.commons.lang3.ArrayUtils;
 
+import io.quarkus.test.configuration.PropertyLookup;
+
 public class ProdLocalhostQuarkusApplicationManagedResource extends LocalhostQuarkusApplicationManagedResource {
 
     private static final String JAVA = "java";
     private static final String QUARKUS_ARGS_PROPERTY_NAME = "quarkus.args";
+    private static final String ENABLE_PREVIEW = "--enable-preview";
+    private static final PropertyLookup JAVA_ENABLE_PREVIEW = new PropertyLookup("ts.enable-java-preview", "false");
 
     private final ProdQuarkusApplicationManagedResourceBuilder model;
 
@@ -25,6 +29,9 @@ public class ProdLocalhostQuarkusApplicationManagedResource extends LocalhostQua
         String[] cmdArgs = extractQuarkusArgs(systemProperties);
         if (model.getArtifact().getFileName().toString().endsWith(".jar")) {
             command.add(JAVA);
+            if (JAVA_ENABLE_PREVIEW.getAsBoolean()) {
+                command.add(ENABLE_PREVIEW);
+            }
             command.addAll(systemProperties);
             command.add("-jar");
             command.add(model.getArtifact().toAbsolutePath().toString());


quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
```diff
@@ -25,6 +25,9 @@ protected List<String> prepareCommand(List<String> systemProperties) {
         String[] cmdArgs = extractQuarkusArgs(systemProperties);
         if (model.getArtifact().getFileName().toString().endsWith(".jar")) {
             command.add(JAVA);
+            if (isEnabledPreview()) {
+                command.add("--enable-preview");
```
michalvavrik, line 29: pls make `--enable-preview` constant

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
```diff
@@ -51,4 +54,14 @@ private String[] extractQuarkusArgs(List<String> systemProperties) {
 
         return args;
     }
+
+    private boolean isEnabledPreview() {
```
michalvavrik, line 58: this method can be static

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
```diff
@@ -51,4 +54,14 @@ private String[] extractQuarkusArgs(List<String> systemProperties) {
...
+        String enablePreviewString = System.getProperty("--enable-preview");
```
michalvavrik, line 59: I don't have strong opinion here, but if this is something specific for this framework. it should be framework config property. Therefore instead of `-D--enable-preview ` create property like `ts.enable-java-preview-features` (please name it yourself) and use `PropertyLookup` and document it when this get merged in Wiki.

Let's here from others though, this is just my preference.

michalvavrik (changes requested):
There is one thing I don't understand - if we only enable preview feature with system properties, how are tests going to pass if the feature is not enabled. I support your main objective is https://github.com/quarkus-qe/quarkus-jdkspecifics/tree/main ? I wonder if it is really desirable to add preview specific tests, we have enough on our plate with testing of what is actually in JDK. If this is only for your personal testing, can't you instead execute it manually, e.g. package and then run it as any other app? 

It's possible I'm missing something, please let me know.

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
rsvoboda: +1 for something like `ts.enable-java-preview-features` instead of `-D--enable-preview`

Should be documented in https://github.com/quarkus-qe/quarkus-test-framework/wiki/2.-Configuration once merged.

rsvoboda (changes requested):
This feature is specific for baremetal use case in prod mode / when application is started using `java -jar`.
It's OK-ish, we do not have a need for this in Kube/OpenShift environment or CLI / DEV mode at this moment.

We could consider more generic approach to allow definition of java command arguments, but could be a bit problematic for Kube/OpenShift and dev mode.

mjurc (commented):
Generally, I agree with points made by @rsvoboda . I don't mind having it here, but maybe a more general approach could be used.

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
jedla97: Updated

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
jedla97: I play with it little and change it to the framework property.

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/ProdLocalhostQuarkusApplicationManagedResource.java
jedla97: Removed as I using the `PropertyLookup` so this method is no longer needed

jedla97:
I play with this a little. At first I was thought that without `-D--enable-preview` the devmode won't be working and fail when running the test containing some preview feature. In my surprise there was no problem running it as it probably can take `--enable-preview` set in pom (this case in jdkSpecific TS).

Generally as this probably only quarkus-qe use case the framework property is better, but as just user who can possibly run it everywhere I would prefer `-D--enable-preview` as it propagated to all usecases/enviroments and it's defined in Java. Just my POV after thinking about it.

michalvavrik:
> I play with this a little. At first I was thought that without `-D--enable-preview` the devmode won't be working and fail when running the test containing some preview feature. In my surprise there was no problem running it as it probably can take `--enable-preview` set in pom (this case in jdkSpecific TS).
> 
> Generally as this probably only quarkus-qe use case the framework property is better, but as just user who can possibly run it everywhere I would prefer `-D--enable-preview` as it propagated to all usecases/enviroments and it's defined in Java. Just my POV after thinking about it.

I'll leave review to others as said, but if you want something propagated to all usecase/environments, can't you use `JDK_JAVA_OPTIONS` set in failsafe plugin? I take it we also need it during build so that is probably the issue, I won't dig into it...

```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-failsafe-plugin</artifactId>
    <version>${failsafe-plugin.version}</version>
    <configuration>
        <environmentVariables>
            <JDK_JAVA_OPTIONS>--enable-preview</JDK_JAVA_OPTIONS>
        </environmentVariables>
    </configuration>
</plugin>
```

jedla97:
> can't you use JDK_JAVA_OPTIONS set in failsafe plugin?

Basically yes in jdk-specific `--enable-preview` it set on multiple places as part of `compilerArgs`, `jvmArgs` or `argLine` and the locally it's used and executed without problem. Not tested it with Kube/OpenShift.

rsvoboda (approved):

mjurc (approved):

PR #904, opened 2023-10-12
Fix nightly failures

fedinskiy:
### Summary

Before that change we used quarkus CLI properties as environment variables in openshift manifests (which we should not have do) and for some reason it worked (which it should not have do). When this behaviour was changed upstream, our jobs started to fail, this change should fix that.

See https://github.com/quarkusio/quarkus/issues/36359 for details

Please check the relevant options

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Dependency update
- [ ] Refactoring
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update
- [x] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [ ] Example scenarios has been updated / added
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #915, opened 2023-10-17
Fix OpenShiftAmqpAmqIT by changing service name and bump ActiveMQ Artemis Broker to same as Dev Svcs are using

michalvavrik:
### Summary

- service name seems to be changed since 3.5.0.CR1 from amq-broker-amqp to amq-amqp. I didn't find a commit that caused it yet, but I'm clear that it works with 3.4.3 and that oc get service says amq-amqp and this fixes failing OC test
- we are using old image version, let's use the same version as dev services are using

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/examples/amq-amqp/src/main/resources/application.properties
+++ b/examples/amq-amqp/src/main/resources/application.properties
@@ -1,4 +1,4 @@
-amqp-host=amq-broker-amqp
+amqp-host=amq-amqp
 amqp-port=5672
 # Configure the AMQP connector to write to the `prices` address
 mp.messaging.outgoing.generated-price.connector=smallrye-amqp
--- a/quarkus-test-service-amq/src/main/java/io/quarkus/test/services/AmqContainer.java
+++ b/quarkus-test-service-amq/src/main/java/io/quarkus/test/services/AmqContainer.java
@@ -13,7 +13,7 @@ import io.quarkus.test.services.containers.model.AmqProtocol;
 @Retention(RetentionPolicy.RUNTIME)
 public @interface AmqContainer {
 
-    String image() default "quay.io/artemiscloud/activemq-artemis-broker:0.1.2";
+    String image() default "quay.io/artemiscloud/activemq-artemis-broker:1.0.22";
 
     String expectedLog() default "Artemis Console available";
 


michalvavrik:
run tests

michalvavrik:
Ad OCP JVM failures - Strimzi failure is known to me and I'm about to work on it tomorrow.

michalvavrik:
Oracle JVM failure is can't be related.

michalvavrik:
@mjurc is busy with 3.2, @fedinskiy please have a look when you can

fedinskiy:
>service name seems to be changed

We need to push upstream folks to document this. @michalvavrik , could you look into that?

fedinskiy:
There is a debug failure as well (In GH CI). Is that expected?

michalvavrik:
> There is a debug failure as well (In GH CI). Is that expected?

I can't tell anything from logs, I'll debug it now, but really, it can't have anything to do with changes this PR is making :-)

michalvavrik:
> > service name seems to be changed
> 
> We need to push upstream folks to document this. @michalvavrik , could you look into that?

I'm working on it, it's not trivial to determine what is it that changed, I already tried for 2 hours yesterday.

michalvavrik:
what has changed is that in 3.5.0.CR1 it for some reason matters what is set in amq.host property in app properties even though we override it with system property. I still can't connect it to any commit. will keep digging. anyway, this PR is right, service name was always as it sets

michalvavrik:
@fedinskiy https://github.com/quarkusio/quarkus/issues/36549, can we merge this now?

michalvavrik:
I guess when the issue is fixed we can put it back, but this time with a comment. it took me while to realize that service name is incorrect, I wonder if that was intention.

michalvavrik:
I wonder if we should set OpenShift environment variables like amq-host instead of amq_host as defined in https://download.eclipse.org/microprofile/microprofile-config-2.0-M1/microprofile-config-spec.html#default_configsources.env.mapping

fedinskiy:
@michalvavrik we should no, this is a problem with our property-to names pipelines, which was fixed here: https://github.com/quarkus-qe/quarkus-test-framework/pull/904/files

I presume, we need to process amqp properties the same way we process quarkus ones, so this bug should be fixed in OpenShiftClient, not in application properties

michalvavrik:
> @michalvavrik we should no, this is a problem with our property-to names pipelines, which was fixed here: https://github.com/quarkus-qe/quarkus-test-framework/pull/904/files
> 
> I presume, we need to process amqp properties the same way we process quarkus ones, so this bug should be fixed in OpenShiftClient, not in application properties

This can only be bug if you ever intended to set incorrect name in app properties, in which case it should be noted in comment. It's just wasting of my time on this setting there something that is not supposed to work.

I'm going to fix it in FW then, ok.

michalvavrik:
This wasn't against you @fedinskiy , I just find it annoying to set there something that is not expected for usage, but has opposite intention without documentation.

fedinskiy:
@michalvavrik I huist noticed, that I made a typo in my answer 😅

We can have anything we want in properties, but for environment variables, we should transfer it into proper format. before 3.5, Quarkus was able to handle both formats, but it has changed since then

PR #941, opened 2023-10-29
Skip Quarkus build done my maven plugin in modules where we experienced race on Windows to find out if its related

michalvavrik:
### Summary

So far we experienced race on Windows in native mode in 3 example modules: `consul`, `management`, `blocking-reactive-modul`. Please see daily build failures. 

While in CI we see failures pretty regularly, I couldn't reproduce it in hundred of attempt (maybe related to Windows machine resources). I presume the root of issue is in Quarkus, but can't really be sure if it is not related to fact that Quarkus Maven plugin builds native executable and right after this, we build it again from our framework (for there are build time properties detected).

- I propose to skip Quarkus build performed by plugin completely in all modes / operating systems in `examples-blocking-reactive` and `management` as this executable is actually never used, we build a new one.
- For `examples/consul` I propose to only skip it in Windows, because there at least once is executable created by plugin used (not that it matter IMO).
- `examples-blocking-reactive` is now _temporarily_ enabled till it fails again (for I want to see if it also fails when the plugin is skipped

Judging by the nature of the failure (race for file that is not created by the plugin), we are going to have it disable in next days when we experience failure, but it is already failing every other day and this skipping was suggested here: https://github.com/quarkusio/quarkus/issues/27061#issuecomment-1715525060. I want to be able to prove skipping plugin build will make no difference.

UPDATE: my current plan is to keep it failing and introduce new code in next days that will print out information on processes involved in race.

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -353,7 +353,7 @@ jobs:
       - name: Build in Native mode
         shell: bash
         run: |
-          mvn -B -fae -s .github/mvn-settings.xml clean install -Pframework,examples,native,skip-tests-on-windows-in-native -Dquarkus.native.container-build=false -Dquarkus.platform.version="${{ matrix.quarkus-version }}"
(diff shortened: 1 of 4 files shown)


mjurc (approved):
Ok, let's try it.

PR #949, opened 2023-11-09
Build Windows native executable by Quarkus Maven plugin

michalvavrik:
### Summary

We experience file access race on Windows during native build. So far it only happens when build is done directly by our FW via `io.quarkus.bootstrap.app.QuarkusBootstrap`. Related code is too complex, debugging would take me weeks.

From now on, we build native executable on Windows always via Quarkus Maven plugin. Whether custom build is required or not. This way we are building executable in a same manner as users will, which makes this approach better in my eyes and we might want to use it in other scenarios in the future when this solution has baked. I've tested it locally with Linux as well.

One palpable difference is that we need to properly propagate those of config properties that are build-time properties (as build is run in a new process without having system properties propagated), hence the `quarkus.native.container-build` change.

On Windows there can be an issue with long paths in case we have many nested project modules, but I think we can handle it with enabling long paths in case we experience it (which I don't expect because I shortened resulting path and QE TS Windows native runs just couple of tests).

Tested in #950.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/LocalhostQuarkusApplicationManagedResource.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/LocalhostQuarkusApplicationManagedResource.java
@@ -16,7 +16,6 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
-import org.junit.jupiter.api.condition.OS;
 
 import io.quarkus.test.bootstrap.Protocol;
 import io.quarkus.test.logging.FileServiceLoggingHandler;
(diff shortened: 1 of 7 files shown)


michalvavrik:
@rsvoboda I expect #950 to pass because first time CI was green except for one module where I had problems with long paths which I already solved. Please have a look so that we can see if this helped to mitigate daily build issue.

michalvavrik:
Made tiny change (one word) for future (to make re-building native executable in same application folder work in case of restart).

rsvoboda:
Commits are not squashed

rsvoboda (approved):
LGTM. Some heavy work was done in `QuarkusMavenPluginBuildHelper`!
Let's wait for CI here and on #950

PR #971, opened 2023-12-01
Use Quarkus Maven plugin to build native executables

michalvavrik:
### Summary

Fact that (so far) #949 fixed win race issue when building native executable on Windows tells me that by using `QuarkusBootstrap` class directly we just hope our setup is same as Quarkus Maven plugin is using. Users will have native executable built by Quarkus maven plugin, and so should we.

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/examples/amq-amqp/pom.xml
+++ b/examples/amq-amqp/pom.xml
@@ -34,4 +34,20 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>${quarkus.platform.group-id}</groupId>
+                <artifactId>quarkus-maven-plugin</artifactId>
+                <version>${quarkus.platform.version}</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>build</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
 </project>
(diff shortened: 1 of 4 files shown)


michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
I think remaining OCP native failures goes down to the fact that `maven.repo.local` needs to be propagated. It should be easy to do, I'll do it and test it.

michalvavrik:
Remaining system properties are propagated anyway, so `maven` specific args are special-case.

michalvavrik:
Ah, or I decided to not propagate non-Quarkus system props I think, yeah, still should be that, logs, images and `ts` props.

michalvavrik:
The safest way would be to propagate whole previous command, I'll check if I can somehow determine that.

michalvavrik:
I've checked - it is possible to do this safely via `ProcessHandle` API, I'll propagate exactly `-D` used in the original command.

michalvavrik:
Okay, I think I got it. We can probably make it more robust in the future, but I think it will work for current use cases. Let's see.

michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
@rsvoboda all green now :-)

rsvoboda (approved):
LGTM!

PR #976, opened 2023-12-10
Update Kafka utils to use images from Maven properties

gtroitsk:
### Summary

KafkaRegistry and KafkaVendor now using image defined in Maven properties. It helps track versions from one place.

Please check the relevant options

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Dependency update
- [x] Refactoring
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update
- [x] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [ ] Example scenarios has been updated / added
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #983, opened 2023-12-16
Revert "Release 1.4.0.Beta3"

gtroitsk:
Reverts quarkus-qe/quarkus-test-framework#978 as [update of Kafka utils](https://github.com/quarkus-qe/quarkus-test-framework/pull/976) turned out to be a bad decision. It is not possible to retrieve systemProperties from FW parent POM in the Quarkus QE TS

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -1,4 +1,4 @@
 name: Quarkus QE Test Framework
 release:
-  current-version: 1.4.0.Beta3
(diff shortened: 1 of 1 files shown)


michalvavrik:
This wouldn't work (as release is out there already and next time we would try to release publishing would probably fail), in any case it is unnecessary, just tell dependabot to skip this verison. Thanks

PR #984, opened 2023-12-17
Revert "Update Kafka utils to use images from Maven properties"

gtroitsk:
Reverts quarkus-qe/quarkus-test-framework#976

This update turned out to be a bad decision. It is not possible to retrieve systemProperties from FW parent POM to use in Quarkus QE TS, properties are ignored.


michalvavrik (approved):
I was wrong to suggest doing this at the first place, apologies, thanks for the revert.

PR #996, opened 2023-12-21
Allow to disable Linux containers detection execution condition as it can provide false positives when Docker host strategy is used and no CLI is available

michalvavrik:
### Summary

Camel Quarkus QE does not install Docker on (some of?) their Windows machines, nor they start Docker on machines that only run OpenShift tests (this part is strange, because condition is only run for `@QuarkusScenario` but I'd need reproducer and time to be worried about OCP), therefore check for Docker CLI can give false positives when only DOCKER_HOST is set, but there is no Docker CLI installed.

My suggestion is that unless Camel Quarkus QE really needs this feature, we only add an option to disable this condition evaluation on Windows. We added this condition to be able to run tests on Windows when Docker is available and at the still time to run TS on Windows when Docker is not available.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:


llowinge (approved):
LGTM, thanks for proactive solution.

PR #1021, opened 2024-01-29
Re-use native executable when test is re-run by FailSafe plugin

michalvavrik:
### Summary

Right now if you do `mvn verify` (or with `clean` but there is Quarkus maven plugin in your POM file) and there are no build-time properties, the native executable from `/target` is re-used. However when there are build-time properties, it is not re-used. Logic behind that is probably _build-time properties requires new build of native executable_. But what if you added a new class, or removed one, or changed source code, doesn't it require new native executable? That is why user should use `clean` like `mvn clean verify` when he made changes that require new executable/JAR (we are talking user changes, not changes done by the test itself).

This way, we can re-use custom builds kept in module target for test re-run by Failsafe plugin when test fails (which saves time and we do not need to test indempotency of Quarkus build). On the other hand it means if there is 10 different test classes with build properties, there will be 10 different native executable kept in target (=we will keep custom builds). Native executable should be small, so I do not consider it an issue.

It saves a lot of time when test is re-run by the Failsafe plugin.

Another effect of this PR is that it makes also native executable build by this framework without build-time properties re-usable, which should allow us to disable Quarkus plugin build in native mode in the future. If we want to . I'll propose and explain that in a separate PR.

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/LocalhostQuarkusApplicationManagedResource.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/LocalhostQuarkusApplicationManagedResource.java
@@ -120,7 +120,7 @@ public abstract class LocalhostQuarkusApplicationManagedResource extends Quarkus
     @Override
     public void restart() {
         stop();
-        if (model.containsBuildProperties()) {
+        if (model.buildPropertiesChanged()) {
             model.build();
         }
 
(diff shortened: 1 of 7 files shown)


michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
run tests

mjurc (approved):
Nothing obviously wrong here, lgtm. Thanks for the update.

PR #1024, opened 2024-01-30
Always skip native build by Quarkus maven plugin to avoid unused native executable builds when all tests require custom builds

michalvavrik:
### Summary

https://github.com/quarkus-qe/quarkus-test-framework/pull/1021 follow-up

1. When all test classes require custom build, the executable build by the Quarkus Maven plugin is ignored. We hope that we mention all of these cases and disable the plugin build, but it's really impossible to keep this in mind when doing every single change in tests.
There is no advantage on always building native executable at the moment when we are yet to know whether we are going to need it or not. Our FW is now using very same maven plugin to build the executable and also supports re-using of executables that do not require custom classes, dependencies or configuration properties.
Strictly speaking, this PR also enables plugin build in JVM in cases it previously wasn't, but I suppose it doesn't matter, I think mostly it was me who added these properties and I was just lazy to create specific native profile or I forgot. it can have positive impact as well (re-using build?).
Unlike JVM build, the native build takes significant time.

2. I mentioned when native build done by the framework is not a custom one, the runner name is `-runner`. It doesn't cause issues on Linux (not sure on Windows as it's not part of the PR workflow, probably, but it needs to be fixed anyway). Sorry. New name will be `quarkus-runner` or `quarkus-runner.exe`.

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/examples/blocking-reactive-model/pom.xml
+++ b/examples/blocking-reactive-model/pom.xml
@@ -9,10 +9,6 @@
     </parent>
     <artifactId>examples-blocking-reactive</artifactId>
     <name>Quarkus - Test Framework - Examples - Blocking + Reactive Model</name>
-    <properties>
-        <!-- Skip build since our framework will build own executable anyway as it detects build-time properties -->
-        <quarkus.build.skip>true</quarkus.build.skip>
-    </properties>
     <dependencies>
         <dependency>
             <groupId>io.quarkus</groupId>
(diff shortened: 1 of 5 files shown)


michalvavrik:
run tests

michalvavrik:
Interesting, this seems essentially wrong as we need to filter out runtime properties as well, which makes it wrong strategy. However it helped me to uncover 2 bugs, which I'll fix in a separate PR.

PR #1047, opened 2024-02-19
Propagate quarkus OPTS to openshift deployment

### LINKED ISSUE #1034: Openshift build deployment strategy does not propagate QUARKUS_OPTS
Using OpenShiftDeploymentStrategy.Build in `@OpenShiftScenario` ignores quarkus-opts properties from properties file, like:
`quarkus.openshift.env.vars.quarkus-opts`
Deployment in openshift has always empty QUARKUS_OPTS as set in https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-openshift/src/main/resources/quarkus-build-openshift-template.yml#L94

These options should be propagated to the openshift deployment.

mocenas:
### Summary

Fix #1034 
Propagate quarkus OPTS to openshift deployment.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Dependency update
- [ ] Refactoring
- [X] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:

### DIFF
--- a/quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/BuildOpenShiftQuarkusApplicationManagedResource.java
+++ b/quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/BuildOpenShiftQuarkusApplicationManagedResource.java
@@ -6,9 +6,12 @@ import static io.quarkus.test.services.quarkus.model.QuarkusProperties.QUARKUS_N
 import static java.util.regex.Pattern.quote;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
 import org.opentest4j.AssertionFailedError;
@@ -18,6 +21,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.quarkus.test.configuration.PropertyLookup;
 import io.quarkus.test.logging.Log;
 import io.quarkus.test.utils.Command;
+import io.quarkus.test.utils.PropertiesUtils;
 
 public class BuildOpenShiftQuarkusApplicationManagedResource
         extends TemplateOpenShiftQuarkusApplicationManagedResource<ArtifactQuarkusApplicationManagedResourceBuilder> {
@@ -26,6 +30,7 @@ public class BuildOpenShiftQuarkusApplicationManagedResource
 
     private static final String QUARKUS_OPENSHIFT_TEMPLATE = "/quarkus-build-openshift-template.yml";
     private static final String IMAGE_TAG_SEPARATOR = ":";
+    private static final String QUARKUS_OPENSHIFT_OPTS_PROPERTY = "quarkus.openshift.env.vars.quarkus-opts";
 
     private String builtImageName = "no-image-yet";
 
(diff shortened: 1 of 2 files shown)


mocenas:
run tests

fedinskiy (commented):
Hmm, this is strange. 
1. application.properties is included in the build and so it should be used anyway. If it is not included, a lot more things would be affected.
2. Why do you need to pass `quarkus.openshift.env.vars.quarkus-opts` to a pod, which already on openshift? From  the name of it, it should be passed to the build made my quarkus-openshift extension, and not by us
3. The only thing, which mentions this property is that bug issue[1] created by us. Is it even exists?

[1] https://github.com/quarkusio/quarkus/issues/29642

mocenas:
> Hmm, this is strange.
> 
>     1. application.properties is included in the build and so it should be used anyway. If it is not included, a lot more things would be affected.
> 
>     2. Why do you need to pass `quarkus.openshift.env.vars.quarkus-opts` to a pod, which already on openshift? From  the name of it, it should be passed to the build made my quarkus-openshift extension, and not by us
> 
>     3. The only thing, which mentions this property is that bug issue[1] created by us. Is it even exists?
> 
> 
> [1] [quarkusio/quarkus#29642](https://github.com/quarkusio/quarkus/issues/29642)

Yes QUARKUS_OPTS really does exists and is effective on native. Where `quarkus.openshift.env.vars.quarkus-opts`  is ineffective and needs to be passed to the pod this way. Basically `quarkus.openshift.env.vars.quarkus-opts` should be translated/transfered into QUARKUS_OPTS. 

See [Josef's investigation about the issue|https://github.com/quarkusio/quarkus/issues/29642#issuecomment-1355041068) about that.

There is a big difference if we deploy the build using openshift-extension, or using our own Build. (see [deployment strategies](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-openshift/src/main/java/io/quarkus/test/scenarios/OpenShiftDeploymentStrategy.java). If we use openshift-extension then `quarkus.openshift.env.vars.quarkus-opts` is handled by it and it is used to set QUARKUS_OPTS.

But using `OpenShiftDeploymentStrategy.Build` (default)  makes it to be deployed to openshift via test-framework, where QUARKUS_OPTS is handled manually by us. And prior to this PR, it was always set to empty (see changes in template).

mocenas:
Openshift failure is unrelated, executor failed mid tests. I will rerun the tests.

mocenas:
run tests

quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/BuildOpenShiftQuarkusApplicationManagedResource.java
```diff
@@ -137,4 +144,15 @@ private String getS2iImage() {
         return model.getContext().getOwner().getProperty(s2iImageProperty.getPropertyKey())
                 .orElseGet(() -> s2iImageProperty.get(model.getContext()));
     }
+
+    private String getQuarkusOpts() {
+        Path applicationProperties = model.getComputedApplicationProperties();
+        if (Files.exists(applicationProperties)) {
+            Map<String, String> properties = PropertiesUtils.toMap(applicationProperties);
+            if (properties.containsKey(QUARKUS_OPENSHIFT_OPTS_PROPERTY)) {
```
fedinskiy, line 152: Maybe getting the value and then checking for `null` will be more clear.

fedinskiy (changes requested):
This action is a little bit surprising from user's standpoint (since we access property, used by `quarkus-openshift`, while not using the extension).

I am not sure we should do it, but we definitely need to log this action (maybe even with content of `QUARKUS_OPTS` property)

quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/BuildOpenShiftQuarkusApplicationManagedResource.java
mocenas: That IMHO depends on programming style. For me, this is more clear.

fedinskiy (approved):
Thank you!

PR #1065, opened 2024-03-16
Support and test Kafka authentication with SASL_SSL protocols

gtroitsk:
### Summary

Add test support for Kafka authentication with SASL_SSL protocols
Tests  `kafka.security.protocol=SASL_SSL`  property

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/examples/kafka/src/test/java/io/quarkus/qe/StrimziKafkaWithDefaultSaslSslMessagingIT.java
+++ b/examples/kafka/src/test/java/io/quarkus/qe/StrimziKafkaWithDefaultSaslSslMessagingIT.java
@@ -1,0 +1,47 @@
+package io.quarkus.qe;
+
+import java.time.Duration;
+
+import org.apache.http.HttpStatus;
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.bootstrap.KafkaService;
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.KafkaContainer;
+import io.quarkus.test.services.QuarkusApplication;
+import io.quarkus.test.services.containers.model.KafkaProtocol;
+import io.quarkus.test.services.containers.model.KafkaVendor;
+
+@QuarkusScenario
+public class StrimziKafkaWithDefaultSaslSslMessagingIT {
+
+    private final static String SASL_USERNAME_VALUE = "client";
+    private final static String SASL_PASSWORD_VALUE = "client-secret";
+    private static final String TRUSTSTORE_FILE = "strimzi-server-ssl-truststore.p12";
+
+    @KafkaContainer(vendor = KafkaVendor.STRIMZI, protocol = KafkaProtocol.SASL_SSL, kafkaConfigResources = TRUSTSTORE_FILE)
+    static final KafkaService kafka = new KafkaService();
+
+    @QuarkusApplication
+    static final RestService app = new RestService()
+            .withProperty("kafka.bootstrap.servers", kafka::getBootstrapUrl)
+            .withProperty("kafka.security.protocol", "SASL_SSL")
+            .withProperty("kafka.ssl.truststore.location", TRUSTSTORE_FILE)
+            .withProperty("kafka.ssl.truststore.password", "top-secret")
+            .withProperty("kafka.ssl.truststore.type", "PKCS12")
+            .withProperty("kafka.sasl.mechanism", "PLAIN")
+            .withProperty("kafka.sasl.jaas.config", "org.apache.kafka.common.security.plain.PlainLoginModule required "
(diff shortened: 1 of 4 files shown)


gtroitsk:
We need to fix failures in FW using snapshot, because it fails after a minute run. It can be caused by bulk dependencies renaming in Quarkus

examples/kafka/src/test/java/io/quarkus/qe/StrimziKafkaWithDefaultSaslSslMessagingIT.java
```diff
@@ -0,0 +1,47 @@
...
+import io.quarkus.test.services.QuarkusApplication;
+import io.quarkus.test.services.containers.model.KafkaProtocol;
+import io.quarkus.test.services.containers.model.KafkaVendor;
+
+@QuarkusScenario
+public class StrimziKafkaWithDefaultSaslSslMessagingIT {
+
+    private final static String SASL_USERNAME_VALUE = "client";
+    private final static String SASL_PASSWORD_VALUE = "client-secret";
+    private static final String TRUSTSTORE_FILE = "strimzi-server-ssl-truststore.p12";
```
jcarranzan, line 22: Apart from `TRUSTSTORE_FILE`, do you think would be needed also a `KEYSTORE_FILE` ?

quarkus-test-service-kafka/src/main/resources/strimzi-default-server-sasl-ssl.properties
```diff
@@ -0,0 +1,167 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
jcarranzan, line 1: is this properties file creation needed? I realized that we already have `strimzi-default-server-ssl.properties` and `strimzi-default-server-sasl.properties`.Maybe you could use them instead.

jcarranzan:
Hi @gtroitsk , I left a couple of comments, but in general LGTM.

examples/kafka/src/test/java/io/quarkus/qe/StrimziKafkaWithDefaultSaslSslMessagingIT.java
gtroitsk: No need to have it in this test because we use one defined in https://github.com/gtroitsk/quarkus-test-framework/blob/f8988043a5730fd1cf265c38cc83b974fc14b982/quarkus-test-service-kafka/src/main/resources/strimzi-default-server-sasl-ssl.properties#L77

If you want to set a new/custom ts/ks you can pass it to container as was did in https://github.com/gtroitsk/quarkus-test-framework/blob/084327f6d3b8a04442dd18847f21d43b46ba419a/examples/kafka/src/test/java/io/quarkus/qe/StrimziKafkaWithCustomSslMessagingIT.java#L20

quarkus-test-service-kafka/src/main/resources/strimzi-default-server-sasl-ssl.properties
gtroitsk: Yes, it is needed. To support SASL_SSL scenario we need to have listeners and mapping to exactly this protocol. `strimzi-default-server-sasl-ssl.properties` is the combination of both files mentioned by you. We can use only single Kafka server.properties file as I know.

jcarranzan: There are some issues closed related to support multiple properties: 

https://github.com/quarkusio/quarkus/issues/12766
https://github.com/quarkusio/quarkus/issues/9869

On the other hand, I think you could add something like that: 
```
 @QuarkusApplication(properties = "application.properties","ssl.properties") 
static RestService app = new RestService();
```
Could you give it a try to dispel any doubts? 
If is not possible, then we will go ahead with your single properties...

gtroitsk: I meant Kafka config `server.properties` file and it can be only one with all required configuration inside.  You can see how we choose properties file based on protocol for Kafka container in [StrimziKafkaContainerManagedResource#getServerProperties](https://github.com/gtroitsk/quarkus-test-framework/blob/f8988043a5730fd1cf265c38cc83b974fc14b982/quarkus-test-service-kafka/src/main/java/io/quarkus/test/services/containers/StrimziKafkaContainerManagedResource.java#L70)

mjurc:
Removing the backport label as this is already in 1.4.z branch

mjurc:
And I am adding the backport label as it turns out that one of the 1.4.2.Beta8 tags actually held 1.4.3.Beta8 release of framework, and I based the 1.4.z branch on that wrongly...

jedla97:
Removing backport label as this was part of https://github.com/quarkus-qe/quarkus-test-framework/pull/1105

PR #1070, opened 2024-03-21
Fix remote DEV tests after breaking package config changes

michalvavrik:
### Summary

Remote DEV mode tests running with Quarkus 999-SNAPSHOT are failing after https://github.com/quarkusio/quarkus/pull/39295 got merged. One example can be https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/8374170247/job/22929729333. Quarkus introduced fallback config interceptors, however we also need to detect it correctly, which seems to be a problem. This PR does:

- disables DEV mode tests in native as there is no point to run them there
- sets new properties and keep the old ones for compatibility as FW is also run with released version

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -436,6 +436,7 @@
                             <execution>
                                 <configuration>
                                     <systemProperties>
+                                        <quarkus.native.enabled>${quarkus.native.enabled}</quarkus.native.enabled>
                                         <native.image.path>${project.build.directory}/${project.build.finalName}-runner</native.image.path>
                                         <quarkus.package.type>${quarkus.package.type}</quarkus.package.type>
                                         <quarkus.native.container-build>${quarkus.native.container-build}</quarkus.native.container-build>
@@ -448,6 +449,8 @@
                 </plugins>
             </build>
             <properties>
+                <quarkus.native.enabled>true</quarkus.native.enabled>
+                <!-- TODO: drop next line when Quarkus is bumped to 3.10 -->
                 <quarkus.package.type>native</quarkus.package.type>
                 <quarkus.native.container-build>true</quarkus.native.container-build>
                 <quarkus.native.native-image-xmx>5g</quarkus.native.native-image-xmx>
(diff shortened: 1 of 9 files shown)


michalvavrik:
Windows failure is fixed here #1074.

pom.xml
```diff
@@ -442,6 +442,7 @@
                             <execution>
                                 <configuration>
                                     <systemProperties>
+                                        <quarkus.native.enabled>${quarkus.native.enabled}</quarkus.native.enabled>
                                         <native.image.path>${project.build.directory}/${project.build.finalName}-runner</native.image.path>
                                         <quarkus.package.type>${quarkus.package.type}</quarkus.package.type>
```
gtroitsk, line 447: Should `quarkus.package.type` here also be dropped after FW use Quarkus 3.10?

quarkus-test-core/src/main/resources/build-time-list
```diff
@@ -70,6 +70,7 @@ quarkus.oidc.enabled
 quarkus.openshift.
 quarkus.otel.enabled
 quarkus.package.type
```
gtroitsk, line 72: Can you please add a comment to remove old property when Quarkus is bumped to 3.10

pom.xml
michalvavrik: this is injection (for lack of better word) of already commented out property, if you drop that property as comment says this will fail so I think it's inherently commented out

quarkus-test-core/src/main/resources/build-time-list
michalvavrik: I didn't add it there on purpose because it's custom format processed by the `io.quarkus.test.services.quarkus.QuarkusApplicationManagedResourceBuilder#isBuildProperty` but after reading the method it shouldn't cause any issues so I'll add it

michalvavrik: done

pom.xml
gtroitsk: That's ok. Thank you for explanation.

quarkus-test-core/src/main/resources/build-time-list
gtroitsk: Good, let's wait for CI to verify

michalvavrik:
@gtroitsk CI is green :-)

gtroitsk (approved):

PR #1088, opened 2024-04-18
Fix Kafka SSL scenarios in FIPS enabled environment by regenerating certs

michalvavrik:
### Summary

We have in place default p12 keystore for Strimzi vendor that has certs generated with algorithm not supported in FIPS. I can't just override it in tests because they all are expecting one truststore. I have decided to re-generate them and refactor it bit to avoid repetition. It's quite quick.

Sadly, this doesn't fix SASL_SSL scenarios as plain auth mechanism just doesn't work in FIPS. I don't know what to do about that part and in the future there will be follow-up, but this PR will help to to fix SSL scenarios at least.

I am adding support for Docker host porting to localhost that we use for Kafka SSL scenarios to also drop repetition.

This PR should be backwards compatible and should not break anything, I run couple of Kafka modules in TS with it.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-service-kafka/pom.xml
+++ b/quarkus-test-service-kafka/pom.xml
@@ -52,5 +52,9 @@
             <optional>true</optional>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>me.escoffier.certs</groupId>
+            <artifactId>certificate-generator</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/examples/kafka/src/test/java/io/quarkus/qe/StrimziKafkaWithDefaultSaslSslMessagingIT.java
+++ b/examples/kafka/src/test/java/io/quarkus/qe/StrimziKafkaWithDefaultSaslSslMessagingIT.java
@@ -17,24 +17,13 @@ import io.quarkus.test.services.containers.model.KafkaVendor;
 @QuarkusScenario
 public class StrimziKafkaWithDefaultSaslSslMessagingIT {
 
-    private final static String SASL_USERNAME_VALUE = "client";
-    private final static String SASL_PASSWORD_VALUE = "client-secret";
-    private static final String TRUSTSTORE_FILE = "strimzi-server-ssl-truststore.p12";
-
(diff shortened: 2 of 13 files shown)


michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
run tests

mjurc:
run tests

quarkus-test-service-kafka/pom.xml
```diff
@@ -52,5 +52,9 @@
             <optional>true</optional>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>me.escoffier.certs</groupId>
+            <artifactId>certificate-generator</artifactId>
```
mjurc, line 57: I'm a bit concerned about this dependency for long term planning, but I guess this is to unblock us short term :)

mjurc (commented):
Besides the one nit, I don't see anything wrong about this. Let's see about the CI.

quarkus-test-service-kafka/pom.xml
michalvavrik: I think it is very useful for generating certs and it's a better way then these certs without README (only some of our truststores and keystores have README with how they were generated and that is provided they are documented right and nothing changed openssl). Clement's dependency is used in testing in Quakrus upstream on many places, except they use JUnit and I used what it depends on (because I don't need annotations).

My idea would be to forget about some certs in resources and start generating certs when SSL is needed everywhere. But ATM I'm only concerned with Kafka.

michalvavrik:
OC jobs are failing but it's not connected to this PR judging by logs. AFAICT I didn't touch anything related to OCP, they have their own builder.

quarkus-test-service-kafka/pom.xml
mjurc: Agreed, let's proceed with using it considering it's also used for upstream Quarkus testing and it is very useful.

mjurc (approved):
Thanks for your efforts.

PR #1109, opened 2024-04-30
Release 1.4.4

mjurc:
### Summary

* Release quarkus-test-framework 1.4.4, which will serve as first stable release for Quarkus 3.8

- [x] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -1,4 +1,4 @@
 name: Quarkus QE Test Framework
 release:
-  current-version: 1.4.2.Beta8
-  next-version: 1.4.3.Beta8
+  current-version: 1.4.4
+  next-version: 1.4.5


.github/project.yml
```diff
@@ -1,4 +1,4 @@
 name: Quarkus QE Test Framework
 release:
-  current-version: 1.4.2.Beta8
-  next-version: 1.4.3.Beta8
+  current-version: 1.4.4
+  next-version: 1.4.5.Beta1
```
rsvoboda, line 4:
```suggestion
  next-version: 1.4.5```

Release plugin adds `-SNAPSHOT`

File: .github/project.yml [hunk 2]
```diff
@@ -1,4 +1,4 @@
 name: Quarkus QE Test Framework
 release:
-  current-version: 1.4.2.Beta8
-  next-version: 1.4.3.Beta8
+  current-version: 1.4.4
```
rsvoboda, line 3: So we are dropping .Final ?

mjurc: I don't mind keeping it in, but considering Quarkus dropped Final too, I thought it would be more clear

.github/project.yml
mjurc: I don't understand this tbh, what is the goal?

.github/project.yml
rsvoboda: OK

.github/project.yml
rsvoboda: "next" imho should be 1.4.5-SNAPSHOT in pom.xml and not `1.4.5.Beta1-SNAPSHOT`
This is how to achieve it

I don't like our current `    <version>1.4.3.Beta11-SNAPSHOT</version>` in https://github.com/quarkus-qe/quarkus-test-framework/blob/main/pom.xml#L6C1-L6C45

mjurc: That version is used just for from-branch builds, it never makes it into any release.

The downside of your approach is that the next person is going to release straight into final, that's my motivation for keeping the Beta in.

You sure you want me to remove it? I don't have a strong opinion either way, but I really don't think we should be telling people to release straight into stable release :D

rsvoboda: We don't need Beta releases on micro releases

mjurc: I've updated it :]

rsvoboda (approved):

Issue #174, opened 2021-08-02
Refactor detection rules of Build properties


Sgitario:
At the moment, the build detection rules are defined in https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-core/src/main/resources/build-time-list.
However, this might change depending on the Quarkus version and may vary over time.
We can overwrite this file in a test suite, however, I think it makes more sense to have some more smart logic here to do what is being done in https://quarkus.io/guides/all-config.html

michalvavrik:
I just hit issue https://github.com/quarkus-qe/quarkus-test-framework/pull/829 due to the fact we have this hardcoded. IMHO it can be as easy as building SmallRye config with correctly set config sources.

michalvavrik:
Note for myself: when this get fixed, we need to revert changes made here https://github.com/quarkus-qe/quarkus-test-suite/pull/1317

PR #1128, opened 2024-05-14
Detect build-time config properties dynamically

michalvavrik:
### Summary

closes: #174

This change detects build-time config property keys in a same fashion as Quarkus does.
Before this PR, we had hard-coded incomplete list of build-time properties which is hard to maintain consistently.
We need to detect build-time properties properly because it tells us when we need custom native executable build and when we do not need it.

There are cons as well:

- loading hardcoded list from a file system is faster than build whole build-time config and config read result
- even though this is public class and Roberto suggested that I use it for this purpose, we can't really call it API, if one day they will refactor it heavily they can just force us to make changes before we can rerun tests with the snapshot (aka close coupling with Quarkus but not through API)
- I didn't find a reasonable way to make this config build silent, in case there is an error (like when expansion fails when there is no such a config property - see `platform.quarkus.native.builder-image` and https://github.com/quarkusio/quarkus/blob/1f5c36fda1a88975b5b791c09194b886dcc3be59/core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java#L251 ) FW build fails because there is no real mitigation that would help; in such a case, it would be a bug in FW that would need to be addressed

Still, it feels like a right thing to do as we can't manage this list manually. It is rather easy to detect if things go wrong - if we do not detect that custom build is needed then test that required it should fail (or it is incorrectly written).

**OpenShift** I run selected tests from my workstation as `run tests` doesn't work ATM and I didn't think it's worth it to create some job manually.

- [ ] Dependency update
- [x] Refactoring
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:


QuarkusQE:
Can one of the admins verify this patch?

mjurc:
run tests

mjurc (commented):
I think this looks OK, but I would really like to see an OCP run :)

mjurc:
run tests

mjurc:
run tests

mjurc:
run tests

michalvavrik:
I looked to the `Run OpenShift Tests in JVM` and it rather seems like a flakyness. In run number 170 the same test that is now failing did not fail. It fails over `Error: ImagePullBackOff`.

mjurc (approved):
LGTM, thanks

Issue #414, opened 2022-02-21
Automated release notes for all streams/branches


mjurc:
Currently, the action used for drafting release notes only works on the main branch.

We need to investigate a way to enable automated release note drafting for the framework.

PR #1159, opened 2024-06-05
Automatically publish release notes

michalvavrik:
### Summary

closes: #414

I used GH CLI to test deletion of API tag (so that we can identify the latest draft), but I didn't try release itself for I don't want to release things. It seems clear enough from docs reference https://cli.github.com/manual/gh_release_edit. I'll try to do actual release and find a way to debug it locally if it fails.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,3 +60,10 @@ jobs:
           branch: ${{github.base_ref}}
           github_token: ${{ secrets.GITHUB_TOKEN }}
           tags: true
+      - name: Publish release notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          gh release edit release-draft --verify-tag --latest --draft=false --target ${{github.sha}} -t ${{steps.metadata.outputs.current-version}} --tag ${{steps.metadata.outputs.current-version}}
+          gh api --method DELETE -H "Accept: application/vnd.github+json" repos/{owner}/{repo}/git/refs/tags/release-draft
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -22,6 +22,6 @@ categories:
     label: 'chore'
 change-template: '- #$NUMBER: $TITLE (@$AUTHOR)'
 template: |
-  ## 1.0.z Changes
+  ## Changes
   $CHANGES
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -60,3 +60,10 @@ jobs:
           branch: ${{github.base_ref}}
           github_token: ${{ secrets.GITHUB_TOKEN }}
           tags: true
+      - name: Publish release notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          gh release edit release-draft --verify-tag --latest --draft=false --target ${{github.sha}} -t ${{steps.metadata.outputs.current-version}} --tag ${{steps.metadata.outputs.current-version}}
```
mjurc, line 68: Beta releases shouldn't be marked as latest, they should be marked as pre-release. They're not release quality, they're beta, and we're not releasing any latest releases from the main branch

mjurc (commented):
In general, is this something we want to do? We should automate the annoying part of the releases (like tagging and pushing to sonatype), but in general, I think this change is going to move people even further from what happens in the releases they produce.

The other thing is that even without release drafter, github produces release notes draft between selected tags already, but I think that people not knowing what happened in their releases and just pushing the button will make things not better

.github/workflows/release.yml
michalvavrik: ok

michalvavrik:
> In general, is this something we want to do? We should automate the annoying part of the releases (like tagging and pushing to sonatype), but in general, I think this change is going to move people even further from what happens in the releases they produce.

it was your idea...

> The other thing is that even without release drafter, github produces release notes draft between selected tags already, but I think that people not knowing what happened in their releases and just pushing the button will make things not better

I don't see how having to remember to go to release notes, click on tag, change title and click on publish is anything but monkey business. I don't read it.

michalvavrik:
okay, if by https://github.com/quarkus-qe/quarkus-test-framework/issues/414 you mean just having it work for non-main branch stream, it's fine with me

michalvavrik:
> We can still have it in btw if people think it's useful, I was just voicing my concerns

I don't mind either way. I was angry with myself spending time on it because I misread #414. Honestly I don't care. Let's forget about it.

PR #1270, opened 2024-08-31
Support OpenShift certificate serving used with TLS registry

michalvavrik:
### Summary

Adding support for certificate serving so that we can test this section of the Quarkus TLS registry extension reference: https://quarkus.io/guides/tls-registry-reference#utilizing-openshift-serving-certificates

I have only implemented it for template-based builder. I think it should be technically possible to test this with extension-based builder, but probably need to configure Quarkus OpenShift extension specific properties inside of the test. It is not really relevant as I care about TLS registry reference verification ATM.

Regarding limitation to Quarkus runtime inside of this PR. I have added it so that I avoid _(not necessarily) possible_ side-effects. I am just implementing what I need and if someone needs more, they need to extend functionality.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:


michalvavrik:
run tests

michalvavrik:
run tests

quarkus-test-openshift/src/main/java/io/quarkus/test/openshift/utils/OpenShiftPropertiesUtils.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    public static URILike getInternalServiceHttpsUrl(ServiceContext ctx) {
+        var serviceName = ctx.getOwner().getName();
+        var projectName = ctx.<OpenShiftClient> get(CLIENT).project();
+        var host = "%s.%s.svc.cluster.local".formatted(serviceName, projectName);
+        return new URILike(Protocol.HTTPS.getValue(), host, getInternalHttpsPort(ctx), "");
+    }
+
+    private static PropertyToValue getPropertyToValue(String propertyValue, String prefix, String separator, String what) {
```
gtroitsk, line 168:
```suggestion
    private static PropertyToValue getPropertyToValue(String propertyValue, String prefix, String separator, String subject) {
```
or something similar

quarkus-test-core/src/main/java/io/quarkus/test/security/certificate/CertificateBuilderImp.java
```diff
@@ -3,7 +3,8 @@
 import java.util.List;
 import java.util.Objects;
 
-record CertificateBuilderImp(List<Certificate> certificates) implements CertificateBuilder {
+record CertificateBuilderImp(List<Certificate> certificates,
```
gtroitsk, line 6: There is a typo Imp -> Impl

quarkus-test-openshift/src/main/java/io/quarkus/test/openshift/utils/OpenShiftPropertiesUtils.java
```diff
@@ -0,0 +1,184 @@
+package io.quarkus.test.openshift.utils;
```
gtroitsk, line 1: It is not related to the class but I give a comment here. Can you please openshift from `openshift/utils` path. I thinks that utils relate to OC is clear from the module name.

quarkus-test-openshift/src/main/java/io/quarkus/test/openshift/utils/OpenShiftPropertiesUtils.java
```diff
@@ -0,0 +1,184 @@
...
+        var configMapNameToAnnotation = getPropertyToValue(propertyValue, ANNOTATED_CONFIG_MAP_PREFIX,
+                ANNOTATED_CONFIG_MAP_SEPARATOR, "annotated-configmap");
+        var annotationKeyToValue = getPropertyToValue(configMapNameToAnnotation.value(), "",
+                ANNOTATED_CONFIG_ANNOTATION_SEPARATOR, "annotation-configmap-annotation");
+        return new PropertyToObj(configMapNameToAnnotation.key(), annotationKeyToValue);
+    }
+
+    public static boolean checkPodReadinessWithStatusInsteadOfRoute(ServiceContext ctx) {
+        // using HTTPS & internal service URL & no exposed route => need to trust readiness
+        return isServingCertificateScenario(ctx) && ServingCertificateConfig.get(ctx).addServiceCertificate();
```
gtroitsk, line 151:
```suggestion
        return isServingCertificateScenario(ctx) && Objects.requireNonNull(ServingCertificateConfig.get(ctx)).addServiceCertificate();
```

quarkus-test-openshift/src/main/java/io/quarkus/test/openshift/utils/OpenShiftPropertiesUtils.java
michalvavrik: no, because I disagree with you. but the fact that we use `io.quarkus.test` packages in different modules is wrong. it's called split packages and it should be avoided. `io.quarkus.test` is telling me nothing about the fact this is class from openshift module. just because someone keeps using `io.quarkus.test` doesn't mean I shouldn't try to do things right way.

quarkus-test-openshift/src/main/java/io/quarkus/test/openshift/utils/OpenShiftPropertiesUtils.java
michalvavrik: it is not null, so I don't need to require it, otherwise `isServingCertificateScenario` would return false. if I applied suggestion it would help IDE with static analysis, but let's hope they will get smarter instead of adding unnecessary code.

quarkus-test-core/src/main/java/io/quarkus/test/security/certificate/CertificateBuilderImp.java
michalvavrik: fixed, thanks

quarkus-test-openshift/src/main/java/io/quarkus/test/openshift/utils/OpenShiftPropertiesUtils.java
michalvavrik: applied, yep, that's better

quarkus-test-openshift/src/main/java/io/quarkus/test/openshift/utils/OpenShiftPropertiesUtils.java
michalvavrik: well, truth be told I am not quite sure if you don't need automatic module name to be same for it to be called split packages, but I simply don't like reusing same package name in different modules.

gtroitsk (approved):
Great work, thank you!

PR #1294, opened 2024-09-09
Added isRHBQ method to check if the version contain string redhat

TimonCzech:
### Summary

Added a new method called is isRHBQ that replaces `QuarkusProperties.getVersion().contains("redhat")` and replaces it's usage

Fixes :#1215

Please check the relevant options

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Dependency update
- [x] Refactoring
- [ ] Release (follows conventions described in the [RELEASE.md](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/RELEASE.md))
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [x] Example scenarios has been updated / added
- [x] Methods and classes used in PR scenarios are meaningful
- [x] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1303, opened 2024-09-12
Move RHBQPlatformAppManager to framework

mocenas:
### Summary

Move RHBQPlatformAppManager from TS to framework.
Add check that created app is indeed using RHBQ not community build. So we can catch this problem early and not after it tries to update community version to RHBQ or something like it.

Also add a factory method to determine which appManager to use. Since the logic between appManagers is growing more complex IMHO it should be in a framework.

- [ ] Dependency update
- [X] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-cli/src/main/java/io/quarkus/test/util/QuarkusCLIUtils.java
+++ b/quarkus-test-cli/src/main/java/io/quarkus/test/util/QuarkusCLIUtils.java
@@ -50,10 +50,9 @@ public abstract class QuarkusCLIUtils {
     public static IQuarkusCLIAppManager createAppManager(QuarkusCliClient cliClient,
             DefaultArtifactVersion oldVersionStream,
             DefaultArtifactVersion newVersionStream) {
-        String quarkusPropertiesVersion = QuarkusProperties.getVersion();
-        if (quarkusPropertiesVersion != null && quarkusPropertiesVersion.contains("redhat")) {
+        if (QuarkusProperties.isRHBQ()) {
             return new RHBQPlatformAppManager(cliClient, oldVersionStream, newVersionStream,
-                    new DefaultArtifactVersion(quarkusPropertiesVersion));
+                    new DefaultArtifactVersion(QuarkusProperties.getVersion()));
         }
         return new DefaultQuarkusCLIAppManager(cliClient, oldVersionStream, newVersionStream);
     }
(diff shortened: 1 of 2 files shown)


quarkus-test-cli/src/main/java/io/quarkus/test/util/QuarkusCLIUtils.java
```diff
@@ -46,6 +47,17 @@ public abstract class QuarkusCLIUtils {
      */
     private static final int GAV_FIELDS_LENGTH = 3;
 
+    public static IQuarkusCLIAppManager createAppManager(QuarkusCliClient cliClient,
+            DefaultArtifactVersion oldVersionStream,
+            DefaultArtifactVersion newVersionStream) {
+        String quarkusPropertiesVersion = QuarkusProperties.getVersion();
+        if (quarkusPropertiesVersion != null && quarkusPropertiesVersion.contains("redhat")) {
```
michalvavrik, line 54: As courtesy of @TimonCzech we now have https://github.com/quarkus-qe/quarkus-test-framework/pull/1294, let's use it. Also, I can see you are doing `quarkusPropertiesVersion != null`, I can't tell you if there is a scenario where it is not null, but as it is public method without contract you should probably add the null check to the utility method as well.

michalvavrik: hmm, on second thought this wasn't backported, so either use it and backport #1294  or not

quarkus-test-cli/src/main/java/io/quarkus/test/util/RHBQPlatformAppManager.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public QuarkusCliRestService createApplicationWithExtraArgs(String... extraArgs) {
+        return assertIsUsingRHBQ(super.createApplicationWithExtraArgs(extraArgs));
+    }
+
+    private QuarkusCliRestService assertIsUsingRHBQ(QuarkusCliRestService app) {
+        try {
+            assertTrue(getQuarkusAppVersion(app).toString().contains("redhat"),
+                    "Created quarkus application does not use redhat version, while should be RHBQ");
```
michalvavrik, line 53:
```suggestion
                    "Created Quarkus application does not use Red Hat version, while should be RHBQ");
```

michalvavrik: hmm, you probably refer to `redhat` literal, in that case if you refer to actual suffix, I'd mention it

michalvavrik (approved):
LGTM; I have comments, but they are not vital

quarkus-test-cli/src/main/java/io/quarkus/test/util/QuarkusCLIUtils.java
mocenas: > hmm, on second thought this wasn't backported, so either use it and backport #1294 or not

If we have this nice method, it would be good to use it, that copy this logic everywhere. I've added backport label to #1294 .

PR #1311, opened 2024-09-13
Allow CLI utils to work with dependencies in CSV

mocenas:
### Summary

While testing big amounts of dependencies to update, it is bad to store them all in code. Allow for storing these in CSV.

I intentionally didn't add support for properties (which we test in similar way) since those are not simple pairs like dependencies and would require more complex solution.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:
- [X] Example scenarios has been updated / added

### DIFF
--- a/examples/quarkus-cli/src/test/java/io/quarkus/qe/QuarkusCliUtilsExampleIT.java
+++ b/examples/quarkus-cli/src/test/java/io/quarkus/qe/QuarkusCliUtilsExampleIT.java
@@ -1,6 +1,10 @@
 package io.quarkus.qe;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
 import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -8,6 +12,7 @@ import jakarta.inject.Inject;
 
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.model.Dependency;
+import org.apache.maven.surefire.shared.lang3.tuple.Pair;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


mocenas:
> Let's have it closer to the place of usage, ie in the testsuite. In that case we also do not need to do two backports and a separate release

@michalvavrik WDYT? As we already have a lot of update-related stuff in FW. Should we keep it rather in FW or move it to TS?

michalvavrik:
> > Let's have it closer to the place of usage, ie in the testsuite. In that case we also do not need to do two backports and a separate release
> 
> @michalvavrik WDYT? As we already have a lot of update-related stuff in FW. Should we keep it rather in FW or move it to TS?

I add stuff like this to FW. Truth be told, I when it comes to opinion like these I simply follow what reviewer wants unless I have strong opinion. I don't think there is one right answer and I want to have work done.

For example, I definitely added a lot of stuff related to TLS and CONFIG commands for Quarkus CLI to FW and I doubt they will be used outside of TS tests. But from that resulted some common abstract CLI result test that I find useful and reused. 

FWIW I think this PR is ok.

mocenas:
> Let's have it closer to the place of usage, ie in the testsuite. In that case we also do not need to do two backports and a separate release

Honestly, I think it would maybe be better to have stuff like this in a TS, but since we already have these support things in FW I don't want to split it. For that reason I would put it in a FW. But thats just my opinion.

BTW: IMHO in it not necessary not backport this, since 3.9 is only quarkus version which has so big renames, that it is worth to use CSV for it.

michalvavrik:
I'll leave it to @fedinskiy . Had my say.

examples/quarkus-cli/src/test/java/io/quarkus/qe/QuarkusCliUtilsExampleIT.java
```diff
@@ -27,6 +32,7 @@
 public class QuarkusCliUtilsExampleIT {
     private static final DefaultArtifactVersion oldVersion = new DefaultArtifactVersion("3.2");
     private static final DefaultArtifactVersion newVersion = new DefaultArtifactVersion("3.8");
+    private static final Path DEPENDENCIES_CSV = Paths.get("src/test/resources/quarkus315dependencyUpdates.csv");
```
rsvoboda, line 35: Where is this file?

I see only `examples/quarkus-cli/src/test/resources/dependenciesToUpdate.csv`

mocenas: Oh. I fixed this, originally the <skipTests> in this module fooled me.

mocenas:
> That's why I would also like to have all of this in TS (including methods with assertions), but alas it is too late for already merged stuff. :-)

OK, you have a point. I'll move this to TS.

PR #1334, opened 2024-09-24
Fix intermittent issue with test OpenShiftS2iQuickstartUsingDefaultsIT on OCP

jcarranzan:
### Summary

I think I found the cause of the intermittent issues in jenkins job, it seems there are 2 tests running concurrently: `OpenShiftContainerRegistryQuickstartUsingDefaultsIT` and `OpenShiftS2iQuickstartUsingDefaultsIT`, both using `QuickstartUsingDefaultsIT` and they are sharing or overwriting the same container image in the OpenShift image registry, labeled as app:latest.

Reviewing the logs, I noticed that the timestamps of the two tests overlap.

This overlapping of both tests is likely trying to pull or push the same image. 

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/examples/external-applications/src/test/java/io/quarkus/qe/BareMetalQuickstartUsingDefaultsIT.java
+++ b/examples/external-applications/src/test/java/io/quarkus/qe/BareMetalQuickstartUsingDefaultsIT.java
@@ -1,11 +1,15 @@
 package io.quarkus.qe;
 
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+
 import io.quarkus.test.bootstrap.RestService;
-import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.test.scenarios.QuarkusScenario;
 import io.quarkus.test.services.GitRepositoryQuarkusApplication;
 
-@OpenShiftScenario
-public class OpenShiftS2iQuickstartUsingDefaultsIT extends QuickstartUsingDefaultsIT {
+@DisabledOnOs(value = OS.WINDOWS, disabledReason = "Windows does not support long file paths")
+@QuarkusScenario
+public class BareMetalQuickstartUsingDefaultsIT extends QuickstartUsingDefaultsIT {
 
     @GitRepositoryQuarkusApplication(repo = "https://github.com/quarkusio/quarkus-quickstarts.git", contextDir = "getting-started")
     static final RestService app = new RestService();
--- a/examples/external-applications/src/test/java/io/quarkus/qe/QuickstartUsingDefaultsIT.java
+++ b/examples/external-applications/src/test/java/io/quarkus/qe/QuickstartUsingDefaultsIT.java
@@ -2,23 +2,16 @@ package io.quarkus.qe;
 
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.DisabledOnOs;
-import org.junit.jupiter.api.condition.OS;
 
 import io.quarkus.test.bootstrap.RestService;
-import io.quarkus.test.scenarios.QuarkusScenario;
-import io.quarkus.test.services.GitRepositoryQuarkusApplication;
 
-@QuarkusScenario
-@DisabledOnOs(value = OS.WINDOWS, disabledReason = "Windows does not support long file paths")
-public class QuickstartUsingDefaultsIT {
+public abstract class QuickstartUsingDefaultsIT {
 
-    @GitRepositoryQuarkusApplication(repo = "https://github.com/quarkusio/quarkus-quickstarts.git", contextDir = "getting-started")
-    static final RestService app = new RestService();
+    protected abstract RestService getApp();
 
     @Test
     public void test() {
-        app.given()
+        getApp().given()
                 .get("/hello")
                 .then()
                 .statusCode(HttpStatus.SC_OK);
--- a/examples/external-applications/src/test/java/io/quarkus/qe/OpenShiftContainerRegistryQuickstartUsingDefaultsIT.java
+++ b/examples/external-applications/src/test/java/io/quarkus/qe/OpenShiftContainerRegistryQuickstartUsingDefaultsIT.java
@@ -1,9 +1,17 @@
 package io.quarkus.qe;
 
+import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.OpenShiftDeploymentStrategy;
 import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.test.services.GitRepositoryQuarkusApplication;
 
 @OpenShiftScenario(deployment = OpenShiftDeploymentStrategy.UsingContainerRegistry)
 public class OpenShiftContainerRegistryQuickstartUsingDefaultsIT extends QuickstartUsingDefaultsIT {
+    @GitRepositoryQuarkusApplication(repo = "https://github.com/quarkusio/quarkus-quickstarts.git", contextDir = "getting-started")
+    static final RestService app = new RestService();
 
+    @Override
+    protected RestService getApp() {
+        return app;
+    }
 }


jcarranzan:
run tests

jcarranzan:
run tests

rsvoboda:
I do not have cycles to look into this today and I need pto tmr, thus re-assigning to @fedinskiy

examples/external-applications/src/test/java/io/quarkus/qe/OpenShiftS2iQuickstartUsingDefaultsIT.java
```diff
@@ -1,8 +1,27 @@
...
+import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.test.services.GitRepositoryQuarkusApplication;
 
 @OpenShiftScenario
+@DisabledOnOs(value = OS.WINDOWS, disabledReason = "Windows does not support long file paths")
 public class OpenShiftS2iQuickstartUsingDefaultsIT extends QuickstartUsingDefaultsIT {
 
+    @GitRepositoryQuarkusApplication(repo = "https://github.com/quarkusio/quarkus-quickstarts.git", contextDir = "getting-started")
+    static final RestService app = new RestService();
```
fedinskiy, line 17: As far as I understand, this field is still in the `QuickstartUsingDefaultsIT`. Maybe we should apply the same change to the `OpenShiftContainerRegistryQuickstartUsingDefaultsIT` and remove it from the base class?

examples/external-applications/src/test/java/io/quarkus/qe/OpenShiftS2iQuickstartUsingDefaultsIT.java
```diff
@@ -1,8 +1,27 @@
...
+
+    @Test
+    public void test() {
```
fedinskiy, line 20: As far as i understand, this is the same test as in `QuickstartUsingDefaultsIT` and `OpenShiftContainerRegistryQuickstartUsingDefaultsIT`. Maybe we should keep it in the base class and just add and  overload method `getApp()`?

fedinskiy (changes requested):
1. If you consider this pr to be ready, then please remove `Draft` status
2. We've solved similar problems several times already, I suggest to check [this](https://github.com/quarkus-qe/quarkus-test-suite/blob/df35b63367a9eab56cf768ebbe62145873c01643/http/management/src/test/java/io/quarkus/qe/extension/OpenShiftBaseDeploymentIT.java#L12) and especially [this](https://github.com/quarkus-qe/quarkus-test-suite/blob/31a80470953182f0e79ba6a600169a8bdde581f4/config/src/test/java/io/quarkus/ts/configmap/api/server/OpenShiftBaseConfigIT.java#L85) as a source of inspiration.

jcarranzan:
run tests

examples/external-applications/src/test/java/io/quarkus/qe/OpenShiftS2iQuickstartUsingDefaultsIT.java
```diff
@@ -1,8 +1,21 @@
...
 
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+
+import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.OpenShiftScenario;
+import io.quarkus.test.services.GitRepositoryQuarkusApplication;
 
 @OpenShiftScenario
+@DisabledOnOs(value = OS.WINDOWS, disabledReason = "Windows does not support long file paths")
```
michalvavrik, line 11: OpenShift tests are never run on Windows, so this change is surplus to the requirements.

examples/external-applications/src/test/java/io/quarkus/qe/QuickstartUsingDefaultsIT.java
```diff
@@ -7,18 +7,16 @@
 
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
-import io.quarkus.test.services.GitRepositoryQuarkusApplication;
 
 @QuarkusScenario
```
michalvavrik, line 11: By making this abstract, you are stopping this from running on bare metal. But you didn't provide separate test for the baremetal.

michalvavrik (changes requested):
I don't believe this PR can cause anything wrong, but after reading the PR description, I have very hard time to believe the PR description.

@fedinskiy please be sure you verify _2 tests running concurrently: OpenShiftContainerRegistryQuickstartUsingDefaultsIT and OpenShiftS2iQuickstartUsingDefaultsIT_ statement before you approve and merge this PR. Thank you.

My questions / doubts:

1. we never run tests concurrently
2. `they are sharing or overwriting the same container image in the OpenShift image registry, labeled as app:latest.` 
  - they are 2 different tests, therefore they are supposed to run in 2 different OCP projects
  - `OpenShiftS2iQuickstartUsingDefaultsIT` uses OpenShift image registry and has `app:latest`, however it is specific to the OCP project: `image-registry.openshift-image-registry.svc:5000/ts-oiatlkeamc/app:latest` and when inspect is, you can see that `latest` is for `image-registry.openshift-image-registry.svc:5000/ts-oiatlkeamc/app@sha256:e59362757c27480af2280d8365dcbfb6d765fe889748a5441df0820ae732c6ca` --> see, there is that ephemeral `ts-oiatlkeamc`
  - `OpenShiftContainerRegistryQuickstartUsingDefaultsIT` is pushed to the quay.io and there is no OCP IS (it is not pushed to the OCP registry at all), if you try to `oc describe pod` you will see that image is in fact `quay.io/someteam/io.quarkus.qe.openshiftcontainerregistryquickstartusingdefaultsit.app`
3. we have (or used to have) plenty of tests like these, what is different here, is it S2I thingy?
4. I have checked logs of OCP FW JVM nightly and I can't see them running concurrently, how often this happens?
5. `Reviewing the logs, I noticed that the timestamps of the two tests overlap.` that sounds like a bug, do you have links? Please link them in the team Slack as we should not share these links here despite the fact they are behind VPN
6. `This overlapping of both tests is likely trying to pull or push the same image.` but one of the 2 tests is taking image from quay and other from internal OCP registry

I will be really happy to be proved wrong, I want this bug fixed. Thanks

examples/external-applications/src/test/java/io/quarkus/qe/QuickstartUsingDefaultsIT.java
jcarranzan: Ok, I've checked because I thought that test was already in another class and executed in baremetal but it was not, then, in that case, I will create another class with just that test to be executed for baremetal

michalvavrik, line 11: So now you have 2 classes:

- `QuickstartDefaultsIT` that is non-abstract and runs, and has identical test like this applicaiton
- `QuickstartUsingDefaultsIT` that only differs in the fact that it is abstract, but:
  - it is annotated with `@QuarkusScenario` which doesn't make sense
  - it is annotated with `@DisabledOnOs` which doesn't make sense considering you are repeating it where relevant already

So, what about the new class will be subclass baremetal exactly as the OpenShift classes are? It's the same thing

jcarranzan:
run tests

examples/external-applications/src/test/java/io/quarkus/qe/BareMetalQuickstartUsingDefaultsIT.java
```diff
@@ -0,0 +1,17 @@
+package io.quarkus.qe;
+
+import io.quarkus.test.bootstrap.RestService;
+import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.GitRepositoryQuarkusApplication;
+
+@QuarkusScenario
+public class BareMetalQuickstartUsingDefaultsIT extends QuickstartUsingDefaultsIT {
```
fedinskiy, line 8: CI says that this should be disabled on windows

jcarranzan: It's disabled now.

fedinskiy (approved):
But let's wait for the CI to be green

michalvavrik:
> But let's wait for the CI to be green

@fedinskiy have you bothered to read any of my comments https://github.com/quarkus-qe/quarkus-test-framework/pull/1334#pullrequestreview-2329565651? How changes in this PR makes any sense? I honestly wonder why I care spent time reviewing this of logic doesn't matter.

fedinskiy:
@michalvavrik I am basing my decision on the following assumptions:
 1. Jose did the proper investigation and sincerely describes things, which he found
 2. The FW is overcomplicated, so there are behaviors, which we do not consider logical, and so the fix may not appear reasonable either.
 3. This PR will not make things worse (thus the requirement of the CI being green).
 
If merging this helps with the failures then Jose's theory describes the FW correctly and we are good. If not — he will need to find another theory and apply its results.

michalvavrik:
> 1. Jose did the proper investigation and sincerely describes things, which he found

What he describes is not happening, sincere as the explanation may be.

> 3. This PR will not make things worse (thus the requirement of the CI being green).
> 
> If merging this helps with the failures then Jose's theory describes the FW correctly and we are good. If not — he will need to find another theory and apply its results.

He describes _This overlapping of both tests is likely trying to pull or push the same image._ is logical nonsense because they are 2 different image registries (OCP vs quay) and _Reviewing the logs, I noticed that the timestamps of the two tests overlap_ is also not happening (I have asked for any log at all, the log Jose provided showed they aren't) .

I don't see any argument, but have it your way. Only reason why I wasted my time on this PR is that:

- I thought that description is either logical
- or this PR hides some serious problem we will run into later and spend time on it again

Have it your way, let's merge it.

rsvoboda:
@jcarranzan Is there a series of green Jenkins runs (10+) which would reassure this has positive impact on the stability?

PR #1351, opened 2024-10-01
Set memory limit for Keycloak container

### LINKED ISSUE #1350: Set memory limit for Keycloak container
As mentioned in the issue https://github.com/quarkusio/quarkus/pull/43601#issue-2557018933, the memory limit for Keycloak container should be set, otherwise, there's an expected rapid increase of memory usage. For more information, see the mentioned issue/PR.

Even for the Quarkus test framework, the memory limit should be set, otherwise OOM can be seen - as reported here: https://github.com/quarkusio/quarkus/issues/43630.

mabartos:
Closes #1350

### Summary

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

----------------------------------------------------------

I'd propose increasing the memory limit (to 1GiB) for Keycloak instance running in a container for the test suite, as it might be more memory-consuming than the Keycloak dev service: https://github.com/quarkusio/quarkus/pull/43601

Container limit is properly used in `quarkus-test-suite` project: 
```
ID            NAME        CPU %       MEM USAGE / LIMIT  MEM %       NET IO        BLOCK IO           PIDS        CPU TIME      AVG CPU %
4521d014fec2  149223041   455.88%     548.9MB / 1.049GB  52.35%      1.5kB / 698B  185.4MB / 1.827MB  56          1m54.708381s  495.01%
```

The solution works as expected for my experiments.

It should close issue: https://github.com/quarkusio/quarkus/issues/43630

But these changes should be probably backported to the `quarkus-test-framework:1.5.x` (or not sure what approach is used in such scenarios).

cc: @jcarranzan

### DIFF
--- a/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/KeycloakContainer.java
+++ b/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/KeycloakContainer.java
@@ -22,5 +22,7 @@ public @interface KeycloakContainer {
 
     String[] command() default {};
 
+    long memoryLimitMiB() default 1000;
+
     Class<? extends ManagedResourceBuilder> builder() default KeycloakContainerManagedResourceBuilder.class;
 }
--- a/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakGenericDockerContainerManagedResource.java
+++ b/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakGenericDockerContainerManagedResource.java
@@ -1,5 +1,7 @@
 package io.quarkus.test.services.containers;
 
+import java.util.Optional;
+
 import org.apache.commons.lang3.StringUtils;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
@@ -38,6 +40,8 @@ public class KeycloakGenericDockerContainerManagedResource extends GenericDocker
         }
 
         container.withCreateContainerCmdModifier(cmd -> cmd.withName(DockerUtils.generateDockerContainerName()));
+        container.withCreateContainerCmdModifier(cmd -> Optional.ofNullable(cmd.getHostConfig())
+                .ifPresent(config -> config.withMemory(convertMiBtoBytes(model.getMemoryLimitMiB()))));
 
         if (isReusable()) {
             Log.info(model.getContext().getOwner(), "Running container on Reusable mode");
@@ -65,5 +69,10 @@ public class KeycloakGenericDockerContainerManagedResource extends GenericDocker
 
     private boolean isPrivileged() {
         return model.getContext().getOwner().getConfiguration().isTrue(PRIVILEGED_MODE);
+    }
+
+    static long convertMiBtoBytes(long valueInMiB) {
+        final var exponentMiB = 20;
+        return (long) (valueInMiB * Math.pow(2, exponentMiB));
     }
 }
--- a/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakContainerManagedResourceBuilder.java
+++ b/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakContainerManagedResourceBuilder.java
@@ -18,6 +18,7 @@ public class KeycloakContainerManagedResourceBuilder extends ContainerManagedRes
     private int restPort;
     private String[] command;
     private String expectedLog;
+    private long memoryLimitMiB;
 
     @Override
     protected String getImage() {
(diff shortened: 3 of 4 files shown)


michalvavrik:
@fedinskiy let's review unless you are busy; if so, pleaes request review from me again; thanks!

michalvavrik:
Thank you @mabartos

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakGenericDockerContainerManagedResource.java
```diff
@@ -66,4 +70,9 @@ protected boolean isReusable() {
     private boolean isPrivileged() {
         return model.getContext().getOwner().getConfiguration().isTrue(PRIVILEGED_MODE);
     }
+
+    private long convertMiBtoBytes(long valueInMiB) {
```
fedinskiy, line 74: Please, add some tests for this method, especially for higher input values.

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/KeycloakContainer.java
```diff
@@ -22,5 +22,7 @@
 
     String[] command() default {};
 
+    long memoryLimitMiB() default 1000;
```
fedinskiy, line 25: I would prefer to provide a memory limit both in units and size (eg `new MemoryLimit(1, Unit.GB)`). Don't we have some utility class in our depedencies for this?

mabartos: I'd prefer it too, but the only suitable type I know of is `MemorySize`, which is part of `io.quarkus:quarkus-core` module. The dependency is not part of the test framework, and I'm afraid to provide some potential circular dependencies.

The MiB memory limit unit for the Keycloak container seems reasonable to me, and I did not want to introduce some duplicate logic around this to achieve such a simple case. 

@fedinskiy Any thoughts?

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakGenericDockerContainerManagedResource.java
mabartos: @fedinskiy Do you have any unit tests for the test framework? Could you point me in some direction? Thanks

jcarranzan: I guess here https://github.com/quarkus-qe/quarkus-test-framework/blob/main/examples/keycloak/src/test/java/io/quarkus/qe/BaseSecurityResourceIT.java could be a place to test your changes.

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/KeycloakContainer.java
fedinskiy: I just checked and  `io.quarkus:quarkus-core` is a dependency of this module (thorough `io.quarkus:quarkus-test-common`), but the MemorySize class is just a wrapper around BigInteger, so it wouldn't help us much. Thanks for looking into it, though

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakGenericDockerContainerManagedResource.java
fedinskiy: Just create a test class in the same module, akin to https://github.com/quarkus-qe/quarkus-test-framework/blob/b300fb3019bba3de21e40669574940440f33d461/quarkus-test-core/src/test/java/io/quarkus/test/services/URILikeTest.java

jcarranzan (commented):
So @mabartos with these modifications it would not be necessary to implement any changes related to `JAVA_OPTS_KC_HEAP="-XX:MaxHeapFreeRatio=30 -XX:MaxRAMPercentage=65"` as suggested here--> https://www.keycloak.org/server/containers#_specifying_different_memory_settings ?
Thanks.

mabartos:
@fedinskiy I added the test.

> So @mabartos with these modifications it would not be necessary to implement any changes related to JAVA_OPTS_KC_HEAP="-XX:MaxHeapFreeRatio=30 -XX:MaxRAMPercentage=65" as suggested here

@jcarranzan it's not suggested. It should only emphasize the approach of setting heap settings via the env var. The default values are `-XX:MaxHeapFreeRatio=70 -XX:MaxRAMPercentage=70` and they should be ok for most deployments.

@jcarranzan Does the documentation and the example seem confusing to you? I can improve the documentation a bit.

fedinskiy (approved):

jcarranzan:
> @fedinskiy I added the test.
> 
> > So @mabartos with these modifications it would not be necessary to implement any changes related to JAVA_OPTS_KC_HEAP="-XX:MaxHeapFreeRatio=30 -XX:MaxRAMPercentage=65" as suggested here
> 
> @jcarranzan it's not suggested. It should only emphasize the approach of setting heap settings via the env var. The default values are `-XX:MaxHeapFreeRatio=70 -XX:MaxRAMPercentage=70` and they should be ok for most deployments.
> 
> @jcarranzan Does the documentation and the example seem confusing to you? I can improve the documentation a bit.

Ok got it, it's not need modify docs, thank you @mabartos

michalvavrik (approved):

PR #1407, opened 2024-11-20
Retry container launch if failed

mocenas:
### Summary

Test containers sometimes fails to launch (especially on podman), we can try to relaunch them in case of a failure.

Edit:
I changed the startup limit count to be configurable. By default it is 1 (will not retry at all), but can be configured using maven CLI `-Dts.global.container-startup-attempts=2`. So we can use it only it jobs where we want to retry.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/DockerContainerManagedResource.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/DockerContainerManagedResource.java
@@ -25,6 +25,7 @@ import org.testcontainers.utility.MountableFile;
 import io.quarkus.test.bootstrap.ManagedResource;
 import io.quarkus.test.bootstrap.Protocol;
 import io.quarkus.test.bootstrap.ServiceContext;
+import io.quarkus.test.configuration.PropertyLookup;
 import io.quarkus.test.logging.Log;
 import io.quarkus.test.logging.LoggingHandler;
 import io.quarkus.test.logging.TestContainersLoggingHandler;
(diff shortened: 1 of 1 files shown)


quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/DockerContainerManagedResource.java
```diff
@@ -37,6 +37,8 @@ public abstract class DockerContainerManagedResource implements ManagedResource
     public static final String DOCKER_INNER_CONTAINER = DockerContainerManagedResource.class.getName() + "_inner";
     private static final String DELETE_IMAGE_ON_STOP_PROPERTY = "container.delete.image.on.stop";
     private static final String TARGET = "target";
+    private static final int CONTAINER_STARTUP_ATTEMPTS = Integer.parseInt(
+            System.getProperty("ts.global.container-startup-attempts", "1"));
```
michalvavrik, line 41: That's not a way we should work with properties. I can give you multiple reasons, for example `ts.global` part is applied for every service, but you may want to set it to just one service with `ts.app.container-startup-attempts`. Admittedly, unlikely, but I can give you other examples, for example it doesn't handle environment variables etc. Even though I expect we will use exactly this system property, I'd prefer if you used `private static final PropertyLookup CONTAINER_STARTUP_ATTEMPTS = new PropertyLookup("container-startup-attempts", "1");`. Thanks

michalvavrik (commented):
PR looks good and thanks very much for looking into it, but please address my comment.

michalvavrik (approved):

PR #1428, opened 2024-12-09
Use 999-SNAPSHOT artifact version for a development

michalvavrik:
### Summary

I'd like to use 999-SNAPSHOT version for a development as it's super annoying to build FW twice (once for development and once for TS development). Quarkus QE TS daily's fails over missing 999-SNAPSHOT https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/12227655427/job/34105123719, I expect this could help, but I didn't look further if it doesn't, I'll find why daily build deploy doesn't work.

- [ ] Dependency update
- [x] Refactoring

### Checklist:

### DIFF
--- a/.github/workflows/daily-deploy.yml
+++ b/.github/workflows/daily-deploy.yml
@@ -26,7 +26,6 @@ jobs:
 
       - name: Maven release 999-SNAPSHOT
         run: |
-          mvn -B versions:set -DnewVersion=999-SNAPSHOT
           mvn -B -DskipTests -DskipITs \
             -DretryFailedDeploymentCount=3 \
             -Psnapshot,framework \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Maven release ${{steps.metadata.outputs.current-version}}
         run: |
           git checkout -b release
-          mvn -B release:prepare -Prelease,framework,examples -DautoVersionSubmodules -DreleaseVersion=${{steps.metadata.outputs.current-version}} -DdevelopmentVersion=${{steps.metadata.outputs.next-version}}-SNAPSHOT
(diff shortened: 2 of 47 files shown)


rsvoboda (commented):
LGTM.

But what about `next-version` in https://github.com/quarkus-qe/quarkus-test-framework/blob/main/.github/project.yml?

`.github/workflows/check-release-version.yml` checks it, but we do not really need it with main. But nor sure how about it when branch is created. Some sanity checks could be needed.

Branches could also switch to `999-SNAPSHOT` based versioning, e.g. `1.5.999-SNAPSHOT`. Wdyt?

michalvavrik:
> LGTM.

Thanks for the review.

> But what about `next-version` in https://github.com/quarkus-qe/quarkus-test-framework/blob/main/.github/project.yml?
> `.github/workflows/check-release-version.yml` checks it, but we do not really need it with main. But nor sure how about it when branch is created. Some sanity checks could be needed.

We still need to release snapshots, so we need to keep track of the current version (although we could infer it from the next version I guess). There can be other consumers of our FW so with Michal we only discussed changing current development version to 999-SNAPSHOT. Feel free to fight back if I am missing something.

> Branches could also switch to `999-SNAPSHOT` based versioning, e.g. `1.5.999-SNAPSHOT`. Wdyt?

Absolutely, I'll follow with it.

michalvavrik:
hey @rsvoboda can you check today? I can also ask for another review if your schedule is tight today. I'd like to deal with it so that I can check dailys with this change. Thanks

rsvoboda (approved):

michalvavrik:
After the struggle with 999-SNAPSHOT release workflow, I don't plan to make same changes in 1.5.z. We can do that later.

rsvoboda:
Let's gain some experience on the main first before battling on many fronts 😇.

PR #1458, opened 2025-01-04
Rely on Java to get the lines, avoid usage of System.lineSeparator()

rsvoboda:
### Summary

Do not use System.lineSeparator(), deployment-build-props.txt file was generated on Unix-like system and causes troubles on Windows systems (in some cases, especially when build time fixed property is redefined). Rely on Java to get the lines instead.

On Windows the whole deployment-build-props.txt was treated as one entry in `buildTimeConfigKeys` before this change. I've added safety check to avoid this situation in the future.

This is the root cause of io.quarkus.ts.http.advanced.ContainerRequestFilterIT and io.quarkus.ts.http.advanced.reactive.ContainerRequestFilterReactiveIT failures on Windows in TS PRs - e.g. https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/12600192714/job/35127050598?pr=2255

Generated applications.properties file was much smaller on Windows than on Linux. The investigation was quite time consuming, but the fix is pretty trivial :)

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
@@ -8,10 +8,10 @@ import static java.util.stream.Collectors.toSet;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -23,6 +23,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import io.quarkus.deployment.configuration.BuildTimeConfigurationReader;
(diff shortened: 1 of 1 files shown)


michalvavrik (approved):
Great work finding it. Change alone is fine, but also:

- this code didn't change since 6/3/24 and while it is a bug and I am surprising short on details, I think it's just a fallback for deployment-module build-time properties. I wonder what has  changed, it is not clear from your PR description
- I think this file can be generated on any system https://github.com/quarkus-qe/quarkus-test-framework/blob/39d58b0dc66a77d564e0531771e4482c2d667979/misc/DeploymentBuildTimeConfigListGeneration.md. What about we keep separating file lines on Java and use `java.nio.file.Files#readAllLines(java.nio.file.Path)`?

rsvoboda:
PR update to rely on Java / commons-io.

> this code didn't change since 6/3/24

I think this got revealed after `Use Quarkus Maven plugin strategy for JVM` changes got merged. That build is "external and uses generated applications.properties.

Execution command doesn't necessarily rely only on the generated properties, some properties are passed through `-D`, for example `22:45:19,866 INFO  Running command: java -Dquarkus.log.console.format=%d{HH:mm:ss,SSS} %s%e%n -Dquarkus.management.enabled=false -Dquarkus.http.port=1103 -jar /Users/rsvoboda/git/quarkus-test-framework/examples/management/target/LocalIT/unmanaged/mvn-build/target/quarkus-app/quarkus-run.jar` for https://github.com/quarkus-qe/quarkus-test-framework/blob/main/examples/management/src/test/java/io/quarkus/qe/LocalIT.java#L26 case

quarkus-test-core/src/main/java/io/quarkus/test/services/quarkus/QuarkusApplicationManagedResourceBuilder.java
```diff
@@ -325,12 +326,15 @@ protected void detectBuildTimeProperties(Map<String, String> computedProperties)
 
             // gather build-time config keys from extensions deployment modules
             // this won't work for named config keys (without regex), but we can tweak that in the future if we need to
-            var deploymentBuildProps = Arrays
-                    .stream(FileUtils
-                            .loadFile("/deployment-build-props.txt")
-                            .split(System.lineSeparator()))
+            var deploymentBuildProps = IOUtils
+                    .readLines(QuarkusApplicationManagedResourceBuilder.class
```
michalvavrik, line 330: I just remembered you could have keep `FileUtils.loadFile("/deployment-build-props.txt")` and just do `.lines().map(String:trim).collect(toSet())`. I knew I forgot some useful method :-). Never mind, this solution is fine.

michalvavrik (approved):

michalvavrik:
Alright, we knew something on Windows was wrong after `Use Quarkus Maven plugin strategy for JVM`, I just didn't know that TS was failing as well. I'll try to remember and re-check next week. Thanks!

rsvoboda:
Daily Windows is not failing because no module is executed :) I noticed that yesterday and I will be looking into it.

michalvavrik:
> Daily Windows is not failing because no module is executed :) I noticed that yesterday and I will be looking into it.

Even better :-D Thanks then!

PR #1422, opened 2024-11-28
Add daily deploy snapshot - change deploy plugin.

mocenas:
### Summary

Follows up on https://github.com/quarkus-qe/quarkus-test-framework/pull/1418.

Fix problems with both daily and release workflows.

Daily was failing on singing, because propagating GPG PASSPHRASE to maven env was missing.

Release was failing, because release workflow has wrong serverId. Server ID in the workflow (which workflow sets up the credentials for) should match the one in pom.xml. I wonder that this worked before.

Also splitting the pom profiles to have separate profile for snapshot and release deploy.
Current release deployement should be the same as before.
Snapshot deployement has successfully run [here](https://github.com/mocenas/quarkus-test-framework/actions/runs/12068258792) 

Please check the relevant options

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Dependency update
- [ ] Refactoring
- [ ] Release (follows conventions described in the [RELEASE.md](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/RELEASE.md))
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update
- [ ] This change requires execution against OCP (use `run tests` phrase in comment)

### Checklist:
- [ ] Example scenarios has been updated / added
- [ ] Methods and classes used in PR scenarios are meaningful
- [ ] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #1463, opened 2025-01-08
Use quarkus-bom snapshot in main

mocenas:
### Summary

This PR continues on https://github.com/quarkus-qe/quarkus-test-framework/pull/1422.

Original goal was to have a daily compatible build of FW and TS. Currently our TS is quarkus 999-SNAPSHOT while framework use 3.17.5. This causes binary incompatibilities between FW and TS. Changing FW to also quarkus 999-SNAPSHOT in main branch.

This might cause instability with daily FW builds but I see no other way to have FW and TS in same line.

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -33,63 +33,6 @@ jobs:
           name: maven-repo-current-fw${{matrix.java}}
           path: maven-repo-current-fw.tgz
           retention-days: 1
-  linux-build-jvm-released:
-    name: Linux - JVM build - Released Versions
-    runs-on: ubuntu-latest
-    needs: validate-format
-    strategy:
-      matrix:
-        quarkus-version: ["current"]
-        java: [ 17 ]
-    outputs:
-      has-flaky-tests: ${{steps.flaky-test-detector.outputs.has-flaky-tests}}
-    steps:
-      - uses: actions/checkout@v4
-      - name: Reclaim Disk Space
-        run: .github/ci-prerequisites.sh
-      - name: Install JDK {{ matrix.java }}
-        uses: actions/setup-java@v4
-        with:
-          distribution: 'temurin'
-          java-version: ${{ matrix.java }}
-          check-latest: true
-          cache: 'maven'
-      - name: Download Maven Repo
-        uses: actions/download-artifact@v4
-        with:
-          name: maven-repo-current-fw${{matrix.java}}
-          path: .
-      - name: Extract Maven Repo
-        shell: bash
-        run: tar -xzf maven-repo-current-fw.tgz -C ~
-      - name: Build in JVM mode
-        run: |
-          mvn -B --no-transfer-progress -fae -s .github/mvn-settings.xml clean install -Pexamples
-      - name: Detect flaky tests
-        id: flaky-test-detector
-        shell: bash
-        if: ${{ hashFiles('**/flaky-run-report.json') != '' }}
-        run: echo "has-flaky-tests=true" >> "$GITHUB_OUTPUT"
-      - name: Rename flaky test run report to avoid file name conflicts
-        id: rename-flaky-test-run-report
-        if: ${{ hashFiles('**/flaky-run-report.json') != '' }}
-        shell: bash
-        run: mv target/flaky-run-report.json target/flaky-run-report-linux-jvm-released.json
-      - name: Archive flaky run report
-        id: archive-flaky-run-report
-        if: ${{ hashFiles('**/flaky-run-report-linux-jvm-released.json') != '' }}
-        uses: actions/upload-artifact@v4
-        with:
-          name: flaky-run-report-linux-jvm-released
-          path: target/flaky-run-report-linux-jvm-released.json
-      - name: Zip Artifacts
-        run: |
-          zip -R artifacts-quarkus${{ matrix.quarkus-version }}-linux-jvm${{ matrix.java }}.zip '*-reports/*'
-      - name: Archive artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: artifacts-quarkus${{ matrix.quarkus-version }}-linux-jvm${{ matrix.java }}
-          path: artifacts-quarkus${{ matrix.quarkus-version }}-linux-jvm${{ matrix.java }}.zip
   linux-build-jvm-latest:
     name: Linux - JVM build - Latest Version
     runs-on: ubuntu-latest
@@ -151,69 +94,6 @@ jobs:
         with:
           name: artifacts-latest-linux-jvm${{ matrix.java }}
           path: artifacts-latest-linux-jvm${{ matrix.java }}.zip
-  linux-build-native-released:
-    name: Linux - Native build - Released Version
-    runs-on: ubuntu-latest
-    needs: validate-format
-    strategy:
-      matrix:
-        quarkus-version: [ "current" ]
-        java: [ 17 ]
-        examples: [
-          'examples/pingpong,examples/restclient,examples/greetings,examples/blocking-reactive-model,examples/https,examples/grpc,examples/consul,examples/infinispan,examples/microprofile,examples/keycloak,examples/kafka,examples/kafka-registry,examples/kafka-streams',
-          '!examples/pingpong,!examples/restclient,!examples/greetings,!examples/blocking-reactive-model,!examples/https,!examples/grpc,!examples/consul,!examples/infinispan,!examples/microprofile,!examples/keycloak,!examples/kafka,!examples/kafka-registry,!examples/kafka-streams'
-        ]
-    outputs:
-      has-flaky-tests: ${{steps.flaky-test-detector.outputs.has-flaky-tests}}
-    steps:
-      - uses: actions/checkout@v4
-      - name: Reclaim Disk Space
-        run: .github/ci-prerequisites.sh
-      - name: Install JDK {{ matrix.java }}
-        uses: actions/setup-java@v4
-        with:
-          distribution: 'temurin'
-          java-version: ${{ matrix.java }}
-          check-latest: true
-          cache: 'maven'
-      - name: Download Maven Repo
-        uses: actions/download-artifact@v4
-        with:
-          name: maven-repo-current-fw${{matrix.java}}
-          path: .
-      - name: Extract Maven Repo
-        shell: bash
-        run: tar -xzf maven-repo-current-fw.tgz -C ~
-      - name: Build
-        run: |
-          mvn -B --no-transfer-progress -fae -s .github/mvn-settings.xml clean install -Pexamples,native -pl '${{ matrix.examples }}'
-      - name: Detect flaky tests
-        id: flaky-test-detector
-        shell: bash
-        if: ${{ hashFiles('**/flaky-run-report.json') != '' }}
-        run: echo "has-flaky-tests=true" >> "$GITHUB_OUTPUT"
-      - name: Rename flaky test run report to avoid file name conflicts
-        id: rename-flaky-test-run-report
-        if: ${{ hashFiles('**/flaky-run-report.json') != '' }}
-        shell: bash
-        run: mv target/flaky-run-report.json target/flaky-run-report-linux-native-released.json
-      - name: Archive flaky run report
-        id: archive-flaky-run-report
-        if: ${{ hashFiles('**/flaky-run-report-linux-native-released.json') != '' }}
-        uses: actions/upload-artifact@v4
-        with:
-          name: flaky-run-report-linux-native-released
-          path: target/flaky-run-report-linux-native-released.json
-      - name: Zip Artifacts
-        if: failure()
-        run: |
-          zip -R artifacts-native-${{ matrix.quarkus-version }}-${{ matrix.java }}.zip '*-reports/*'
-      - name: Archive artifacts
-        if: failure()
-        uses: actions/upload-artifact@v4
-        with:
-          name: artifacts-native-${{ matrix.quarkus-version }}-${{ matrix.java }}
-          path: artifacts-native-${{ matrix.quarkus-version }}-${{ matrix.java }}.zip
   windows-build-jvm-latest:
     name: Windows - JVM build - Latest Version
     runs-on: windows-latest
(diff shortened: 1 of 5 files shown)


mocenas:
cc @mjurc

michalvavrik:
I actually agree with this change @mocenas , maybe we should give @QuarkusQE change to fight back if someone strongly disagree. Anyway, this is goes through, you need to fix GH CI and I don't remember Jenkins, maybe check them as well.

jedla97:
Also if this is merged the https://github.com/quarkus-qe/quarkus-test-framework/blob/main/CONTRIBUTING.md?plain=1#L133 should be updated

rsvoboda (changes requested):
> Also if this is merged the https://github.com/quarkus-qe/quarkus-test-framework/blob/main/CONTRIBUTING.md?plain=1#L133 should be updated

It's a must have

mocenas:
> I actually agree with this change @mocenas , maybe we should give @QuarkusQE change to fight back if someone strongly disagree. Anyway, if this goes through, you need to fix GH CI workflows and I don't remember Jenkins, maybe check them as well.

Is it really necessary to change the GH workflows? I used the same approach we have in TS - add snapshot repos to the parent pom, so pipeline can download quarkus' snapshots from there. IMHO it is much simpler and more manageable this way.

mocenas:
> > Also if this is merged the https://github.com/quarkus-qe/quarkus-test-framework/blob/main/CONTRIBUTING.md?plain=1#L133 should be updated
> 
> It's a must have

I'm not sure what should be changed here. I've added a line to don't forget to change a quarkus version, but AFAIK that should be done anyway when fixing the branch to certain quarkus stream. Otherwise I don't think there is more to add.

jedla97:
> Is it really necessary to change the GH workflows? I used the same approach we have in TS - add snapshot repos to the parent pom, so pipeline can download quarkus' snapshots from there. IMHO it is much simpler and more manageable this way.

Afaik now we run [released version](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/.github/workflows/daily.yaml#L39) and [latest version](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/.github/workflows/daily.yaml#L74) so in my POV there will be need for some adjustment otherwise we will run the same test 2x

rsvoboda:
I think `.github/workflows/daily.yaml` needs a bit more changes.
 - Do we really need `quarkus-main-build` step ? Sonatype snapshots repo is configured .
 - Will `linux-build-jvm-released` be ever executed? almost feels like it should be dropped.

mocenas:
> I think `.github/workflows/daily.yaml` needs a bit more changes.
> 
>     * Do we really need `quarkus-main-build` step ? Sonatype snapshots repo is configured .
> 
>     * Will `linux-build-jvm-released` be ever executed? almost feels like it should be dropped.

I agree we can drop `quarkus-main-build` step It should not be necessary.
I'm not sure about `linux-build-jvm-released`. This step will not be part of normal daily pipelines. IDK if there is some  usecase, where we want to execute it on another branch - there it would make sense.

mocenas:
WDYT now @michalvavrik @rsvoboda ?

mocenas:
I'm OK with dropping "released" sections from daily pipeline.
But my thoughts about keeping it in PR was that it will be (or might be) easier to manage for other branches - it's easier to have one "updated" pipeline and copy to branches. But I don't mind doing it other way.

pom.xml
```diff
@@ -77,6 +77,31 @@
         <reruns>2</reruns>
         <flaky-run-reporter.version>0.1.4</flaky-run-reporter.version>
     </properties>
+    <repositories>
+        <repository>
+            <!-- repository where quarkus 999-snapshot is built -->
+            <id>ossrh</id>
+            <url>https://s01.oss.sonatype.org/content/repositories/snapshots/</url>
```
michalvavrik, line 84: 1. I don't think it is desirable until we have one source of snapshot, when we drop internal nexus for snapshots, this will be ok (I think we already do that in QE TS so this is maybe too strict comment)
2. this change is not really required, we are already running FW with Quarkus snapshot
3. if you insist on this change, please adjust workflows and drop the `quarkus-snapshots-mvn-settings.xml` entirely

.github/workflows/pr.yaml
```diff
@@ -33,10 +33,32 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Install JDK {{ matrix.java }}
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.java }}
+          check-latest: true
+          cache: 'maven'
+      - name: Extract quarkus version
+        run: echo "quarkus-version="$(mvn -q -Dexec.executable=echo -Dexec.args='${quarkus.platform.version}' --non-recursive exec:exec) >> $GITHUB_OUTPUT
```
michalvavrik, line 55: so this `detect-quarkus-version` is useless on main and you have it here only to make creating of new branches workflows easier? I am not a fan, but please at least add comment so that it has clear purpose. thanks

michalvavrik:
If we stop testing released versions like I can see in comments (for example @rsvoboda ) it means we will never test backports. We will never know if backports to 3.17.6 applied without issues because we will never run it.

.github/workflows/pr.yaml
mocenas: I'm not convinced that this is best way But we want to test against released versions in branches. If you think that copying PR workflow from previous branch is better way, I'm not against it. It just seemed clunky to me.

pom.xml
mocenas: I don't think this change is the only way, but IMHO it is the cleanest and simplest way to do it.
IMHO it is best to have this snapshot source setup in pom, that setting it in every workflow. Or is the a reason to have different snapshot source in workflows/environments?

mocenas:
> If we stop testing released versions like I can see in comments (for example @rsvoboda ) it means we will never test backports. We will never know if backports to 3.17.6 applied without issues because we will never run it.

I'm not sure if we understand each other. My impression was to drop testing on released version only on main branch (where released would equal latest - so we run only one of them). But would be kept for other branches (I thought github allows to have different CI for branches but I might be wrong). So testing against released quarkus versions should be done on backport PRs. 
Did I got it wrong?

.github/workflows/pr.yaml
michalvavrik: these are little thing, I don't mind either way. but if you think when creating a new branch you can simply use `main` workflows, I think you are wrong. I have only done it on the TS, but there, I always had to make changes manually. let's not spend too much time discussing this.

pom.xml
michalvavrik: no as far as I am concerned, but it's as I said, there won't be a point to have several snapshot registries, so drop internal nexus snapshot builds as well.

michalvavrik:
> > If we stop testing released versions like I can see in comments (for example @rsvoboda ) it means we will never test backports. We will never know if backports to 3.17.6 applied without issues because we will never run it.
> 
> I'm not sure if we understand each other. My impression was to drop testing on released version only on main branch (where released would equal latest - so we run only one of them). But would be kept for other branches (I thought github allows to have different CI for branches but I might be wrong). So testing against released quarkus versions should be done on backport PRs. Did I get it wrong?

Are you referring to the 1.5.z etc. branches? They doesn't concern the latest released Quarkus version.
There is no daily build for 3.17.6, no PR build for 3.17.6, where is 3.17.6 tested? Now, I don't say that testing it with FW makes super sense, but you are cutting one testing option, so I want:

- people (and you) be aware
- raise awareness that now, we will not testing it at all

pom.xml
michalvavrik: and drop the `quarkus-snapshots-mvn-settings.xml` it won't have a value after your changes

michalvavrik: actually, I didn't  mention it before I think it is just paranoid, but the fact we had our own daily snapshot build meant that:

- we don't need to trust that Quarkus pushes the latest build to the snapshot repo (AKA they build it correctly)
- snapshot registry gives us that build

but I would still drop our snapshot build.

rsvoboda:
That was my impression that TS and TF main are moving fully to 999-SNAPSHOT of Quarkus, no runs with latest Quarkus upstream version. Only the branches would follow the LTS streams.

CONTRIBUTING.md
```diff
@@ -138,6 +138,7 @@ When creating new branch please ensure following items:
  - Pin external application branches, see for example https://github.com/quarkus-qe/quarkus-test-framework/pull/905
  - Pin CLI to concrete stream, see for example https://github.com/quarkus-qe/quarkus-test-framework/pull/918
  - Update GH Actions to use the right Quarkus branch, see for example https://github.com/quarkus-qe/quarkus-test-framework/pull/920
+ - Change quarkus.platform.version property in parent pom.xml to use specific quarkus version, not snapshot
```
michalvavrik, line 141:
```suggestion
 - Change `quarkus.platform.version` property in parent pom.xml to use specific quarkus version, not snapshot
```

.github/workflows/pr.yaml
```diff
@@ -33,10 +33,32 @@ jobs:
...
           path: maven-repo-current-fw.tgz
           retention-days: 1
+  detect-quarkus-version:
+    name: Detect quarkus version
+    runs-on: ubuntu-latest
+    needs: validate-format
+    outputs:
+      quarkus-version: ${{ steps.detect-quarkus-version.outputs.quarkus-version }}
+    strategy:
+      matrix:
```
michalvavrik, line 43: I doubt that Quarkus version will differ per Java version, so the matrix feels surplus to the requirements.

michalvavrik (changes requested):
@mocenas if you are lost in the flood in all these comments, here are actions points required if you want this in merged by me:

- drop settings xml with snapshot repos as we don't need it
- adjust jenkins jobs that are using it so that they don't fail
- drop jenkins job building quarkus and jenkins job building fw (both snapshots deployed to our internal nexus)
- open discussion whether we should test the latest released Quarkus version somewhere because we are loosing this (weak) test coverage
- additional update to the https://github.com/quarkus-qe/quarkus-test-framework/blob/main/CONTRIBUTING.md#branches is required to answer following question: where is `3.15.999-SNAPSHOT`  https://github.com/quarkusio/quarkus/blob/3.15/pom.xml#L10 taken from for the `latest` runs of the PR workflow? it is not deployed to the snapshots repository, so whoever branch the next LTS also needs to add Quarkus build for the `3.x.999-SNAPSHOT`

Apart of that, changes looks good.

mocenas:
> @mocenas if you are lost in the flood in all these comments, here are actions points required if you want this in merged by me:
> 
>     * drop settings xml with snapshot repos as we don't need it
> 
>     * adjust jenkins jobs that are using it so that they don't fail
> 
>     * drop jenkins job building quarkus and jenkins job building fw (both snapshots deployed to our internal nexus)
> 
>     * open discussion whether we should test the latest released Quarkus version somewhere because we are loosing this (weak) test coverage
> 
>     * additional update to the https://github.com/quarkus-qe/quarkus-test-framework/blob/main/CONTRIBUTING.md#branches is required to answer following question: where is `3.15.999-SNAPSHOT`  https://github.com/quarkusio/quarkus/blob/3.15/pom.xml#L10 taken from for the `latest` runs of the PR workflow? it is not deployed to the snapshots repository, so whoever branch the next LTS also needs to add Quarkus build for the `3.x.999-SNAPSHOT`
> 
> 
> Apart of that, changes look good.

Thanks for the wrap up. I will follow this. But I first need to resolve problems in altered daily pipeline. It currently fails the run without the quarkus build.

CONTRIBUTING.md
mocenas: I've applied this change in my commit.

mocenas:
> @mocenas if you are lost in the flood in all these comments, here are actions points required if you want this in merged by me:
> 
>     * drop settings xml with snapshot repos as we don't need it
> 
>     * adjust jenkins jobs that are using it so that they don't fail
> 
>     * drop jenkins job building quarkus and jenkins job building fw (both snapshots deployed to our internal nexus)
> 
>     * open discussion whether we should test the latest released Quarkus version somewhere because we are loosing this (weak) test coverage
> 
>     * additional update to the https://github.com/quarkus-qe/quarkus-test-framework/blob/main/CONTRIBUTING.md#branches is required to answer following question: where is `3.15.999-SNAPSHOT`  https://github.com/quarkusio/quarkus/blob/3.15/pom.xml#L10 taken from for the `latest` runs of the PR workflow? it is not deployed to the snapshots repository, so whoever branch the next LTS also needs to add Quarkus build for the `3.x.999-SNAPSHOT`
> 
> 
> Apart of that, changes look good.

- I kept the snapshot settings.xml as it is used the workflows in the end
- I didn't change jenkins jobs using this file, as it is not removed
- I updated the contribution guide. Now there is a step to copy PR pipeline from previous LTS stream. These contain a jobs to build quarkus main e.g. version 3.15.999-SNAPSHOT. IMHO it is not necessary to describe it more in the guide.

CONTRIBUTING.md
```diff
@@ -137,7 +137,9 @@ When creating new branch please ensure following items:
  - There are no pending PRs for relevant RHBQ or Quarkus stream
  - Pin external application branches, see for example https://github.com/quarkus-qe/quarkus-test-framework/pull/905
  - Pin CLI to concrete stream, see for example https://github.com/quarkus-qe/quarkus-test-framework/pull/918
+ - Use GH Action pr, based on previous stream branch, for example https://github.com/quarkus-qe/quarkus-test-framework/blob/37f099ad8851f05ebdf1a9284dc006c4849b7bf4/.github/workflows/pr.yaml
  - Update GH Actions to use the right Quarkus branch, see for example https://github.com/quarkus-qe/quarkus-test-framework/pull/920
+ - Change `quarkus.platform.version` property in parent pom.xml to use specific quarkus version, not snapshot
```
michalvavrik, line 142:
```suggestion
 - Change `quarkus.platform.version` property in parent pom.xml to use specific Quarkus version, not snapshot
```
but let's not change it unless someone else requests another changes

michalvavrik (approved):
LGTM

mocenas:
> * CI is not running because your changes are invalid https://github.com/quarkus-qe/quarkus-test-framework/actions/runs/12709606389 please fix that

CI is fixed.

rsvoboda (approved):

PR #1467, opened 2025-01-09
Test creating of Quarkus applications and extensions with snapshot CLI targeting both Quarkus snapshot and the latest released version

michalvavrik:
### Summary

Currently by default when we test snapshots (like 999-SNAPSHOT, 3.15.999-SNAPSHOT etc) with explicitly set platform like `--platform-bom=io.quarkus::999-SNAPSHOT` and non-snapshots also with explicitly set platform like `--platform-bom=io.quarkus::999-SNAPSHOT`.

That means that we:

1. don't test CLI without setting Quarkus platform (or stream)
2. don't test whether newer Quarkus CLI (like 999-SNAPSHOT one) can create application with older (e.g. currently latest released version) Quarkus application. this means we could miss issues like this one https://github.com/quarkusio/quarkus/issues/42649

This PR aims to make it easier to test creating both snapshot application/extension and the latest released application/extension. That it, it only comes into effect if the snapshot is tested.

How it works? You know parametrized tests, this is same principle, except we just use our test template provider. This way, we can also disable testing of the "latest released version" when we hit some incompatibility issues between released version and snapshot CLI. We can disable testing of the latest released Quarkus with the snapshot CLI using the `ts.global.cli.snapshot.test-released-quarkus=false` configuration property.

I have no illusion this is perfect, it doesn't cover every scenario we are testing and it can't. We need to explicitly declare what we want to test with both snapshot and the latest released Quarkus version (without setting `-P`) because wiring it implicitly would be too much magic noone would remember.

After this is merged, in the Quarkus QE Test Suite we must manually choose and alter tests that should run with both versions.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/examples/quarkus-cli/src/test/java/io/quarkus/qe/QuarkusCliClientIT.java
+++ b/examples/quarkus-cli/src/test/java/io/quarkus/qe/QuarkusCliClientIT.java
@@ -26,8 +26,10 @@ import org.junit.jupiter.api.Test;
 import io.quarkus.test.bootstrap.QuarkusCliClient;
 import io.quarkus.test.bootstrap.QuarkusCliDefaultService;
 import io.quarkus.test.bootstrap.QuarkusCliRestService;
+import io.quarkus.test.bootstrap.QuarkusVersionAwareCliClient;
 import io.quarkus.test.bootstrap.config.QuarkusConfigCommand;
 import io.quarkus.test.scenarios.QuarkusScenario;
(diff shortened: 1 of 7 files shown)


mjurc (approved):
I think this is better than the current state we have, especially considering the requirement for testing CLI built from Quarkus main against both released registry and snapshot catalog taken from Quarkus main build. I think it's elegant, considering these requirements :D

PR #1474, opened 2025-01-16
Fix test that use Strimzi Kafka containers because enabling of Kraft mode fails over discrepancy between node and broker id

michalvavrik:
### Summary

Literally every Kafka test and Infinispan test that use Strimzi fails to start a Strimzi container https://github.com/quarkus-qe/quarkus-test-suite/actions/runs/12810036907/job/35716160691. I checked dependency tree in `quarkus-test-suite/messaging/kafka-streams-reactive-messaging`:

```
[INFO] +- io.quarkus.qe:quarkus-test-service-kafka:jar:999-SNAPSHOT:test
[INFO] |  +- org.testcontainers:kafka:jar:1.20.4:test
[INFO] |  +- io.strimzi:strimzi-test-container:jar:0.109.1:test
```

Our POM manages Strimzi version we declare explicitly https://github.com/quarkus-qe/quarkus-test-framework/blob/main/pom.xml#L72 and we know that there is something that can override this version, because it is documented: https://github.com/quarkus-qe/quarkus-test-framework/blob/d99146304f52c9a5ccfeede1d423a85374e0bd19/pom.xml#L125. I have tried to enforce Strimzi version we set in our POM, but that must probably be done in TS, not sure if it is important, it works now.

The issue itself is that effective `io.strimzi:strimzi-test-container` version to the `0.109.1` while our explicitly set version `0.107.0`. And in `0.109.1` is validation that says node id must be same as the broker id https://github.com/strimzi/test-container/blob/974411ff0989675b14693493579fcf9516c4a5b4/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java#L657. That is also not a new thing, if you check this javadoc in our FW, it explicitly mentions it: https://github.com/quarkus-qe/quarkus-test-framework/blob/d99146304f52c9a5ccfeede1d423a85374e0bd19/quarkus-test-service-kafka/src/main/java/io/quarkus/test/services/containers/strimzi/ExtendedStrimziKafkaContainer.java#L71. However, we cannot just set `nodeId` and then `brokerId` because if you mention Strimzi code, it doesn't use method argument for the validation, but existing instance broker id `this.useKraft && this.brokerId != this.nodeId`. So if you set correct `withBrokerId`, it still fails because it doesn't use the method argument, it uses `this.brokerId` for validation :-).

Long story short, this PR updates `strimzi-test-container` to match the effective one in Quarkus QE Test Suite and makes sure that we set first broker id and node id before we enabler Kraft mode.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
         <oracle.image>docker.io/gvenzl/oracle-free:23-slim-faststart</oracle.image>
         <postgresql.image>docker.io/postgres:16</postgresql.image>
         <mysql.image>docker.io/mysql:8.4</mysql.image>
-        <strimzi.testcontainers.version>0.107.0</strimzi.testcontainers.version>
+        <strimzi.testcontainers.version>0.109.1</strimzi.testcontainers.version>
         <infinispan.image>docker.io/infinispan/server:15.0</infinispan.image>
         <infinispan-legacy.image>docker.io/infinispan/server:13.0</infinispan-legacy.image>
         <!-- TODO use official image if this fixed https://github.com/hashicorp/docker-consul/issues/184 -->
--- a/quarkus-test-service-kafka/src/main/java/io/quarkus/test/services/containers/strimzi/ExtendedStrimziKafkaContainer.java
+++ b/quarkus-test-service-kafka/src/main/java/io/quarkus/test/services/containers/strimzi/ExtendedStrimziKafkaContainer.java
@@ -74,8 +74,7 @@ public class ExtendedStrimziKafkaContainer extends StrimziKafkaContainer {
      * since config/kraft/server.properties contains node.id=1, we have to use this value
      */
     public ExtendedStrimziKafkaContainer enableKraftMode() {
-        return (ExtendedStrimziKafkaContainer) super.withKraft()
-                .withBrokerId(1);
+        return (ExtendedStrimziKafkaContainer) super.withNodeId(1).withBrokerId(1).withKraft();
     }
 
     public void configureScram(String name, String password) {


michalvavrik:
run tests

michalvavrik:
run tests

michalvavrik:
@jedla97 kindly have a look at your earliest convenience, I'd like to see this in the TS daily build so that I can address (possible) hidden TS failures.

michalvavrik:
Yeah, and @jedla97 I triggered OCP runs, but I didn't realize that this is TC Strimzi, so it shouldn't be related to the OCP. Just FYI.

jedla97 (approved):
Thanks, looks fine to me! Also this problem should be fixed in next release base on https://github.com/strimzi/test-container/issues/124

michalvavrik:
> Thanks, looks fine to me! Also this problem should be fixed in next release base on [strimzi/test-container#124](https://github.com/strimzi/test-container/issues/124)

ah https://github.com/strimzi/test-container/issues/124 , good catch

michalvavrik:
so @fedinskiy already knew about this https://github.com/quarkus-qe/quarkus-test-framework/pull/1456#issuecomment-2582014159, at least we have a workaround in the place (and it seems like OK code to keep to me anyway)

jedla97:
Native ocp failed with `PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested targe`. This is definitely related to this change

michalvavrik:
> Native ocp failed with `PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested targe`. This is definitely related to this change

Some time ago I applied a trick that copies cert bundle from executor machine to the ubi container that is used when s2i build these tests. Maybe something has changed or it is flaky solution, will have to get back to it when there is time....

PR #1527, opened 2025-03-17
Implement support for Keycloak production mode

jedla97:
### Summary

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/examples/keycloak/src/test/java/io/quarkus/qe/BaseSecurityResourceIT.java
+++ b/examples/keycloak/src/test/java/io/quarkus/qe/BaseSecurityResourceIT.java
@@ -5,6 +5,10 @@ import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM_BASE_PATH;
 import static io.quarkus.test.bootstrap.KeycloakService.DEFAULT_REALM_FILE;
 import static org.hamcrest.Matchers.equalTo;
 
+import java.security.KeyManagementException;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.keycloak.authorization.client.AuthzClient;
@@ -19,15 +23,16 @@ public abstract class BaseSecurityResourceIT {
     static final String CLIENT_SECRET_DEFAULT = "test-application-client-secret";
     static final String NORMAL_USER = "test-normal-user";
 
-    @KeycloakContainer(command = { "start-dev", "--import-realm" })
-    static final KeycloakService keycloak = new KeycloakService(DEFAULT_REALM_FILE, DEFAULT_REALM, DEFAULT_REALM_BASE_PATH);
+    @KeycloakContainer(command = { "start", "--import-realm", "--hostname-strict=false" })
+    static final KeycloakService keycloak = new KeycloakService(DEFAULT_REALM_FILE, DEFAULT_REALM, DEFAULT_REALM_BASE_PATH,
+            true);
 
     private AuthzClient authzClient;
 
     protected abstract RestService getApp();
 
     @BeforeEach
-    public void setup() {
+    public void setup() throws NoSuchAlgorithmException, KeyStoreException, KeyManagementException {
         initAuthzClient();
     }
 
@@ -41,7 +46,7 @@ public abstract class BaseSecurityResourceIT {
                 .body(equalTo("Hello, user test-normal-user"));
     }
 
-    private void initAuthzClient() {
+    private void initAuthzClient() throws NoSuchAlgorithmException, KeyStoreException, KeyManagementException {
         authzClient = keycloak.createAuthzClient(CLIENT_ID_DEFAULT, CLIENT_SECRET_DEFAULT);
     }
 
--- a/examples/keycloak/src/test/java/io/quarkus/qe/SecurityResourceIT.java
+++ b/examples/keycloak/src/test/java/io/quarkus/qe/SecurityResourceIT.java
@@ -1,17 +1,24 @@
 package io.quarkus.qe;
 
+import static io.quarkus.test.bootstrap.KeycloakService.KEYSTORE_PASSWORD;
+
 import io.quarkus.test.bootstrap.RestService;
 import io.quarkus.test.scenarios.QuarkusScenario;
+import io.quarkus.test.services.Certificate;
 import io.quarkus.test.services.QuarkusApplication;
 
 @QuarkusScenario
+@Certificate()
 public class SecurityResourceIT extends BaseSecurityResourceIT {
 
     @QuarkusApplication
     static final RestService app = new RestService()
             .withProperty("quarkus.oidc.auth-server-url", keycloak::getRealmUrl)
             .withProperty("quarkus.oidc.client-id", CLIENT_ID_DEFAULT)
-            .withProperty("quarkus.oidc.credentials.secret", CLIENT_SECRET_DEFAULT);
+            .withProperty("quarkus.oidc.credentials.secret", CLIENT_SECRET_DEFAULT)
+            .withProperty("quarkus.oidc.tls.tls-configuration-name", "oidc")
+            .withProperty("quarkus.tls.oidc.trust-store.jks.path", keycloak.getTrustStoreName())
+            .withProperty("quarkus.tls.oidc.trust-store.jks.password", KEYSTORE_PASSWORD);
 
     @Override
     public RestService getApp() {
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/ContainerManagedResourceBuilder.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/ContainerManagedResourceBuilder.java
@@ -25,6 +25,7 @@ public class ContainerManagedResourceBuilder implements ManagedResourceBuilder {
     private String[] command;
     private List<MountConfig> mounts = new ArrayList<>();
     private Integer port;
+    private Integer securedPort;
     private boolean portDockerHostToLocalhost;
 
     protected String getImage() {
@@ -43,6 +44,10 @@ public class ContainerManagedResourceBuilder implements ManagedResourceBuilder {
         return port;
     }
 
+    protected Integer getSecuredPort() {
+        return securedPort;
+    }
+
     protected ServiceContext getContext() {
         return context;
     }
@@ -50,8 +55,13 @@ public class ContainerManagedResourceBuilder implements ManagedResourceBuilder {
     @Override
     public void init(Annotation annotation) {
         Container metadata = (Container) annotation;
-        init(metadata.image(), metadata.command(), metadata.expectedLog(), metadata.port(),
-                metadata.portDockerHostToLocalhost());
+        if (getSecuredPort() == -1) {
+            init(metadata.image(), metadata.command(), metadata.expectedLog(), metadata.port(),
+                    metadata.portDockerHostToLocalhost());
+        } else {
+            init(metadata.image(), metadata.command(), metadata.expectedLog(),
+                    metadata.port(), metadata.securePort(), metadata.portDockerHostToLocalhost());
+        }
         this.mounts = Arrays.stream(metadata.mounts()).sequential()
                 .map(mount -> new MountConfig(mount.from(), mount.to()))
                 .toList();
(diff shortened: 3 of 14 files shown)


jedla97:
run tests

jedla97:
run tests

jedla97:
run tests

jedla97:
run tests

jedla97:
OCP failiures are not caused by this PR. They are also in nightly build

examples/keycloak/src/test/java/io/quarkus/qe/BaseSecurityResourceIT.java
```diff
@@ -19,8 +19,9 @@ public abstract class BaseSecurityResourceIT {
     static final String CLIENT_SECRET_DEFAULT = "test-application-client-secret";
     static final String NORMAL_USER = "test-normal-user";
 
-    @KeycloakContainer(command = { "start-dev", "--import-realm" })
-    static final KeycloakService keycloak = new KeycloakService(DEFAULT_REALM_FILE, DEFAULT_REALM, DEFAULT_REALM_BASE_PATH);
+    @KeycloakContainer(command = { "start", "--import-realm", "--hostname-strict=false" })
```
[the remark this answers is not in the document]
michalvavrik: BTW why is `hostname-strict` a problem on localhost? https://www.keycloak.org/server/all-config says "Should always be set to true in production, unless your reverse proxy overwrites the Host header. If enabled, the hostname option needs to be specified." so there is that "production" thing and I was thinking, is `localhost` an issue? Certs should have correct DNS name.

quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/ContainerManagedResourceBuilder.java
```diff
@@ -43,15 +44,19 @@ protected Integer getPort() {
         return port;
     }
 
+    protected Integer getSecuredPort() {
```
michalvavrik, line 47: Feels like a great deal of this PR is connected to the fact that you decided to use `securedPort` alongside unsecured port, which scenario does require both? I'd expect that if you want to test secured communication, you disable unsecured communication.

quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/DockerContainerManagedResource.java
```diff
@@ -200,4 +200,8 @@ private boolean isResource(String key) {
     private boolean isSecret(String key) {
         return key.startsWith(SECRET_PREFIX);
     }
+
+    public GenericContainer<?> getInnerContainer() {
```
michalvavrik, line 204: I think this shouldn't be necessary, I already made sure you can get `inner container`, you can do it like here https://github.com/quarkus-qe/quarkus-test-suite/blob/3a93e22ab489ff24192c47c6fcf544a8e254799c/sql-db/narayana-transactions/src/test/java/io/quarkus/ts/transactions/MssqlTransactionGeneralUsageIT.java#L37

michalvavrik: but it's fine if you insist on this, just make it `protected`

examples/keycloak/src/test/java/io/quarkus/qe/SecurityResourceIT.java
```diff
@@ -11,7 +13,10 @@ public class SecurityResourceIT extends BaseSecurityResourceIT {
     static final RestService app = new RestService()
             .withProperty("quarkus.oidc.auth-server-url", keycloak::getRealmUrl)
             .withProperty("quarkus.oidc.client-id", CLIENT_ID_DEFAULT)
-            .withProperty("quarkus.oidc.credentials.secret", CLIENT_SECRET_DEFAULT);
+            .withProperty("quarkus.oidc.credentials.secret", CLIENT_SECRET_DEFAULT)
+            .withProperty("quarkus.oidc.tls.tls-configuration-name", "oidc")
+            .withProperty("quarkus.tls.oidc.trust-store.jks.path", keycloak.getTrustStoreName())
```
michalvavrik, line 18: `getTrustStoreName` for `trust-store.jks.path`? Is the name path?

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/bootstrap/KeycloakService.java
```diff
@@ -87,4 +163,71 @@ private String normalizeRealmBasePath(String realmBasePath) {
...
+        }
+    }
+
+    private void setupTruststoreName(Path trustStorePath) throws IOException {
+        if (isBareMetalPlatform()) {
+            trustStoreName = trustStorePath.toAbsolutePath().toString();
+        } else {
+            //Need to copy the truststore to this location to enable some OpenShift Quarkus scenarios to work
+            Files.copy(trustStorePath,
+                    Path.of("target", "classes", trustStorePath.getFileName().toString()).toAbsolutePath(),
```
michalvavrik, line 216: I don't understand this, this will need explanation as what the comment mentions sounds like a bug in FW.

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/bootstrap/KeycloakService.java
```diff
@@ -87,4 +163,71 @@ private String normalizeRealmBasePath(String realmBasePath) {
 
         return realmBasePath;
     }
+
+    private void prepareKeycloakToStartInProdMode(Format keystoreFormat) {
```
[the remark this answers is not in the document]
michalvavrik: I think you will have to add option to specify `SAN` though

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/bootstrap/KeycloakService.java
```diff
@@ -1,39 +1,103 @@
...
      *
      * @param realmFile for example /test-realm-realm.json
      * @param realmName
      * @param realmBasePath such as "/realms" used by Keycloak 18 or "auth/realms" used by previous versions
+     * @param runKeycloakInProdMode the prod mode needs to set up certificate and use https protocol
+     * @param keystoreFormat the format which should be used to enable TLS on Keycloak
      */
-    public KeycloakService(String realmFile, String realmName, String realmBasePath) {
+    public KeycloakService(String realmFile, String realmName, String realmBasePath, boolean runKeycloakInProdMode,
+            Format keystoreFormat) {
```
michalvavrik, line 70: Let's not expose `io.smallrye.certs.Format` into our API (or classes that users will use, call it whatever you want), we should hide this stuff from users as it is impl. detail. You can use `io.quarkus.test.services.Certificate.Format`. If you wonder why is it on the annotation, I think that it is the right place and having `keystoreFormat` in the `KeycloakService` is the wrong one

michalvavrik: These information should come from the context, e.g. `io.quarkus.test.bootstrap.SqlServerService#getTlsProperties()`

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/bootstrap/KeycloakService.java
```diff
@@ -1,39 +1,103 @@
...
         this(realmName);
         this.realmBasePath = normalizeRealmBasePath(realmBasePath);
-        withProperty("KEYCLOAK_REALM_IMPORT", "resource_with_destination::" + REALM_DEST_PATH + "|" + realmFile);
+        this.runKeycloakInProdMode = runKeycloakInProdMode;
+        withProperty("KEYCLOAK_REALM_IMPORT", RESOURCE_WITH_DESTINATION_PREFIX + REALM_DEST_PATH + "|" + realmFile);
+        if (runKeycloakInProdMode) {
+            prepareKeycloakToStartInProdMode(keystoreFormat);
```
michalvavrik, line 76: right, this shouldn't happen on the class instantiation, but in the resource builder when the managed resource is build or when the container is initialized for the baremetal

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/OpenShiftKeycloakContainerManagedResource.java
```diff
@@ -1,8 +1,28 @@
 package io.quarkus.test.services.containers;
 
+import io.quarkus.test.bootstrap.Protocol;
+import io.quarkus.test.services.URILike;
+
 public class OpenShiftKeycloakContainerManagedResource extends OpenShiftContainerManagedResource {
 
+    private static final int HTTP_PORT = 80;
+    private static final int HTTPS_PORT = 443;
```
michalvavrik, line 9: Sorry if it is a stupid question, but how does it work? Here you use `443` for HTPS and on `KeycloakContainer` you have `int securedPort() default 8443;` so when I use it for OCP, it goes from 443 to 8443 somehow?

michalvavrik (commented):
I left some comments, I feel bit uncomfortable asking for so many changes, but I don't like how the certificates and changes in general are done. I'll probably yield if you provide good arguments...

quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/DockerContainerManagedResource.java
jedla97: I make it back private and store it when creating container (`initContainer()`)

examples/keycloak/src/test/java/io/quarkus/qe/SecurityResourceIT.java
jedla97: Good point, renamed

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/bootstrap/KeycloakService.java
jedla97: Now set of cert type is in `@KeycloakContainer` and can be set as `certificateFormat = Certificate.Format.JKS`

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/OpenShiftKeycloakContainerManagedResource.java
jedla97: It's not stupid question, I spend some time to understand why these values are needed. The 80 and 443 is need as they are the ports values which are the apps on openshift exposed. So the openshift `https://keycloak-secured-<project-name>.<ocp-apps-url>` only listen on 443 port and the request is forwarded to keycloak which using 8443. So the KC port is not visible from outside. Atm we using it only for exposing http port 80 in https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-openshift/src/main/java/io/quarkus/test/services/containers/OpenShiftContainerManagedResource.java#L30. This default port is not used if the app is exposed as internal service setign the `openshift.use-internal-service-as-url`

jedla97:
Sorry for long wait @michalvavrik I updated the PR. But looking at https://github.com/quarkus-qe/quarkus-test-framework/pull/1601 that PR doing something similar with exposing the port so I would would like to wait for that PR. And modify my changes how the port is exposed.

> I left some comments, I feel bit uncomfortable asking for so many changes, but I don't like how the certificates and changes in general are done.

No problem at all. The all suggestions are valid!

Can you look mainly if the creating and setting certificate and other things around it looks fine to you now without work around exposing the secure port as I mention I would look at it more when https://github.com/quarkus-qe/quarkus-test-framework/pull/1601 is merged.

quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/OpenShiftKeycloakContainerManagedResource.java
michalvavrik: Got it, thanks.

michalvavrik (commented):
I think this looks good, I still have doubt regarding moving the cert to resources, but it's not a blocker. You will need to coordinate with the other OCP HTTPS PR and ask someone (Fedor/Martin/Michal) to approve this, but what I can see right now looks good.

One minor ask - you have code specific for a PEM certificates. I think you should test it, because PEM can be difficult. Maybe just replace that JKS test, I don't think we need to care about JKS anymore. Despite its usage, it is very much legacy.

jedla97:
> One minor ask - you have code specific for a PEM certificates. I think you should test it, because PEM can be difficult. Maybe just replace that JKS test, I don't think we need to care about JKS anymore. Despite its usage, it is very much legacy.

I'll add test for it

jedla97:
run tests

jedla97:
Hi @michalvavrik this is ready for review. I added the test for PEM certificates

Also I looked at https://github.com/quarkus-qe/quarkus-test-framework/pull/1527#pullrequestreview-2904092599 and think that these two scenarious need it to be part of resources when building the app. I change deployment strategy from `Build` to `UsingOpenShiftExtension` in one of the openshift test in TS which using `@Certificate` on Quarkus deployment. It also failed to find the keystore (forget to check if the was present in openshift project). I not sure if it's a bug in TF or some need for these types of build strategies.

michalvavrik:
Took me quite a time, but I got rid of few things I disliked about this PR:

1. certificates shouldn't be copied to resources, but should be mounted as secrets
2. didn't like trust all strategy for KC AuthZ client
3. storing OpenShift client on static variable instead in the context

Feel free to refactor it whatever way you like, we will need to run whole OpenShift test coverage since I only run tests in `examples/keycloak` and registry scenarios failed on pushing docker tag (probably missed some switch, don't know), let see how tests go.

jedla97:
Thanks for improving it! I think it's look great

Issue #1052, opened 2024-02-22
SSL is not supported on OpenShift


fedinskiy:
See https://github.com/quarkus-qe/quarkus-test-framework/blob/668b87dce594ddb56adcf3cd2288de020eaa972f/quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/OpenShiftQuarkusApplicationManagedResource.java#L103

rsvoboda:
The link in description is invalid, please update it + add link to a "fixed" version, e.g. using tag or commit

fedinskiy:
Current state:

We can support passthrough[1] (="only app certificate, no cluster certificates") ssl for quarkus apps(which is the most important issue) relatively easy:
1. Add 
```
    - name: "https"
      port: 8443 # service port
      targetPort: 8443 # app port
``` 
to `spec.ports` for a service.
2. run command `oc create route passthrough app-https --service=app --port=8443`

The more complicated thing is not do it (especially the first part) when we do not use https. This may require a serious refactor of framework around OpenShiftClient. Other tricky part is integration with management interface (it may support both http and https on the same port).

Alternative we can try to always create http ports, just like we do for management[2]. It may be a little bit simpler.



[1] https://www.redhat.com/en/blog/encryption-secure-routes-openshift
[2] https://github.com/quarkus-qe/quarkus-test-framework/commit/43433b925ea02d8e515188a9bd786489bea4c0a1#diff-647d66ebaa722ac2ebfe71b76774da3139216cd2741daf2bb4ca975eec652059

PR #1601, opened 2025-06-02
Proposal for TLS for OCP services

mocenas:
### Summary

This a proposal for how we could enable TLS in our OCP tests. Currently its just a POC of how it might work. It is far from finished.

I intend to support only creating TLS pass-through routes (TLS is terminated on the app). I don't see a point in doing any other OCP TLS termination types for our use cases.

Core idea is that app annotations will have a new parameter `ocpTlsPort`. In case it is set, FW will create a TLS passthrough route to that app. This will only setup a route, it will not setup the TLS in app or nothing else.

I really want to avoid changing the OCP yaml templates on the fly or any similar stuff. IMHO it would be a huge mess.

I tried two implementation approaches:

1. Currently implemented for keycloak (side effect also for other container based app). Where I currently have 2 OCP yaml templates and based on the `ocpTlsPort` param, it will take the one with TLS or without it. Positive of this is, that we don't create anything else on the OCP when we don't use TLS. Negative is, that we would have to maintain two templates and with any more branching, this can get out of control.
2. Currently for `@QuarkusApplication`. It will just create the TLS port and service every time and based on `ocpTlsPort` it will (not) expose the TLS (create/not the HTTPS route). It is simpler to implement and manage. IMHO create the TLS port on deployment/service while not using it is not that big of a deal. It creates some (possibly) unnecessary stuff, but I don't think 

Another way might be to actually adjust the OCP resources on fly, using OpenShiftClientImpl, but I'm not sure how well that is set up in the FW.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/services/QuarkusApplication.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/services/QuarkusApplication.java
@@ -41,4 +41,11 @@ public @interface QuarkusApplication {
      * Enable GRPC configuration. This property will map the gPRC service to a random port.
      */
     boolean grpc() default false;
+
+    /**
+     * Enable TLS pass-through route on OpenShift. If yes state port to forward TLS traffic to.
+     * If set to 0, TLS will be disabled.
+     * This option does !not! configure the app to setup TLS, it just creates TLS route in OCP.
+     */
+    int ocpTlsPort() default 0;
 }
--- a/quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/TemplateOpenShiftQuarkusApplicationManagedResource.java
+++ b/quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/TemplateOpenShiftQuarkusApplicationManagedResource.java
@@ -71,6 +71,11 @@ public abstract class TemplateOpenShiftQuarkusApplicationManagedResource<T exten
                 this::internalReplaceDeploymentContent,
                 addExtraTemplateProperties(),
                 model.getContext().getServiceFolder().resolve(DEPLOYMENT));
+        if (model.getOcpTlsPort() != 0) {
+            client.createTlsPassthroughRoute(model.getContext().getName(),
+                    model.getContext().getName() + "-tls",
+                    model.getOcpTlsPort());
+        }
     }
 
     private String internalReplaceDeploymentContent(String content) {
@@ -86,6 +91,7 @@ public abstract class TemplateOpenShiftQuarkusApplicationManagedResource<T exten
         content = content.replaceAll(quote("${SERVICE_NAME}"), model.getContext().getName())
                 .replaceAll(quote("${INTERNAL_PORT}"), "" + getInternalPort())
                 .replaceAll(quote("${INTERNAL_INGRESS_PORT}"), "" + ingressInternalPort)
+                .replaceAll(quote("${TLS_PORT}"), "" + model.getOcpTlsPort())
                 .replace("${MANAGEMENT_PORT}", "" + model.getManagementPort());
 
         return replaceDeploymentContent(content);
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/Container.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/Container.java
@@ -28,4 +28,10 @@ public @interface Container {
     Class<? extends ManagedResourceBuilder> builder() default ContainerManagedResourceBuilder.class;
 
     Mount[] mounts() default {};
+
+    /**
+     * Enable TLS pass-through route on OpenShift. If yes state port to forward TLS traffic to.
+     * If set to 0, TLS will be disabled.
+     */
+    int ocpTlsPort() default 0;
 }
--- a/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/ContainerManagedResourceBuilder.java
+++ b/quarkus-test-containers/src/main/java/io/quarkus/test/services/containers/ContainerManagedResourceBuilder.java
@@ -25,6 +25,7 @@ public class ContainerManagedResourceBuilder implements ManagedResourceBuilder {
     private String[] command;
     private List<MountConfig> mounts = new ArrayList<>();
     private Integer port;
+    private Integer ocpTlsPort;
     private boolean portDockerHostToLocalhost;
 
     protected String getImage() {
@@ -43,6 +44,10 @@ public class ContainerManagedResourceBuilder implements ManagedResourceBuilder {
         return port;
     }
 
+    protected Integer getOcpTlsPort() {
+        return ocpTlsPort;
+    }
+
     protected ServiceContext getContext() {
         return context;
     }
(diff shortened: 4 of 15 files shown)


fedinskiy:
This change is more or less aligns with the earlier investigation(https://github.com/quarkus-qe/quarkus-test-framework/issues/1052#issuecomment-2884206588), and it has the same problems as solution #2:
1. A lot of services/routes, even when they are not used (but as you noticed, it may be ok)
2. No support for https on management port (but maybe we do not need that yet)
Also, a brief looks shows, that there is a hardcoded port (8443), so I am not sure how tunable the app would be

mocenas:
> Also, a brief looks shows, that there is a hardcoded port (8443), so I am not sure how tunable the app would be
that port is hardcoded only for communication between service<->route. I don't see a point of having that configurable. So you can still set a port on your application and port outside should still be 433.

I'm not really familiar with management port, can't say if we need TLS on it.

fedinskiy:
@mocenas
> I'm not really familiar with management port, can't say if we need TLS on it.

We do, but it is not urgent:
https://github.com/quarkus-qe/quarkus-test-suite/blob/2b483fa2165907e5d0f18687896915ccab05e07c/http/management/src/test/java/io/quarkus/qe/OpenShiftOptionsIT.java#L12

michalvavrik:
FWIW I'd not merge anything until you have tests for the new functionality in the FW.

quarkus-test-core/src/main/java/io/quarkus/test/services/QuarkusApplication.java
```diff
@@ -41,4 +41,11 @@
      * Enable GRPC configuration. This property will map the gPRC service to a random port.
      */
     boolean grpc() default false;
+
+    /**
+     * Enable TLS pass-through route on OpenShift. If yes state port to forward TLS traffic to.
+     * If set to 0, TLS will be disabled.
+     * This option does !not! configure the app to setup TLS, it just creates TLS route in OCP.
+     */
+    int ocpTlsPort() default 0;
```
fedinskiy, line 50: We already have `ssl()` method above and https port in the Qaurkus app (configured by `quarkus.http.ssl-port`), so maybe we do not need this option.

On the other hand, if this value also means "port on the app", when we can use it with management port as well. Could you try that? We have a test in `examples/management`: https://github.com/quarkus-qe/quarkus-test-framework/blob/626955a767ee705ad066b273c99caf5cc2acc42b/examples/management/src/test/java/io/quarkus/qe/OpenShiftTLSIT.java#L10

mocenas: We can rely on `ssl()` method, whether to create or not the secured route. But IMHO it is good to have the port configurable, as port number may be different for apps.

Yet it means port on the app. But it only creates HTTPS route. But it is also necessary to setup HTTPS certificates etc. Our `@Certificate` annotation currently does not support that. So right not it cannot be easily used in management example. But in theory it should work if we add also support for certificate generation.

mjurc: About `ocpTlsPort` parameter - I don't mind it, but can it be made optional? We already have `ssl = true` for bare metal scenarios. Is there a reason to diverge in behavior for kube/OCP? The way I imagine it is that route's created with arbitrarily picked default port if there's no further specification (e.g. setting `ocpTlsPort`), but otherwise the declarative config of `QuarkusApplication` should stay same-ish not matter if it is bare metal or kubernetes scenario

mjurc:
Some other notes:

1)
> Another way might be to actually adjust the OCP resources on fly, using OpenShiftClientImpl, but I'm not sure how well that is set up in the FW.
I don't really mind using the templates, but it would be nice to keep in mind that creating extra resources can cause side effects on test scenarios. We should try making the tests as focused on actually creating and testing what they use.

Framework's classes (in particularly in openshift module) have a reference to namespaced kubernetes/OpenShift fabric8 client, so the only operation that needs to be done is to create/destroy services as needed, right? Wouldn't this be much easier to implement programmatically than creating two services for everything always? 

2) I'm not sure what the deal with TLS and management port is, actually. Does it really matter for any test use case that if TLS is set up, Quarkus will secure management port with TLS? I just cannot see what the actual issue is here. Maybe @fedinskiy can chime in with an explanation

mocenas:
> Framework's classes (in particularly in openshift module) have a reference to namespaced kubernetes/OpenShift fabric8 client, so the only operation that needs to be done is to create/destroy services as needed, right? Wouldn't this be much easier to implement programmatically than creating two services for everything always?

Yes it would be much better in theory. I don't want to use the word easier. Because as @fedinskiy stated, our OpenShiftClient is complicated and really not setup to do on-the-fly changes. Main problem in this is, that our OpenShiftClient is basically just copy pasting the YAML templates without any semantical understanding of the content.
I will try, if it is possible to do this programatically with reasonable effort. But I don't want to rewrite half of our OpenshiftClient as part of this PR.

mocenas:
I pushed some changes. for `@QuarkusApp` it now relies on `ssl() == true` whether to create TLS route or no.
Also TLS stuff is now patched programatically. It is not that big of change into OCP client. With side effect that deployment and service that is patched will be rolled out once more (in practice it will restart service and app (if already build and deployed)).

michalvavrik:
> It requires to manually change the builder on annotation to @QuarkusApp(builder = OpenShiftQuarkusApplicationManagedResourceBuilder.class. I don't like that, but default builder is hardcoded in the annotation and I need to override builder's step configureCertificates() (default one is setting cert paths on FS that would make the app fail). Any idea how to do this better? (read - how to do it, so we don't need to change the builder).

Look at what `io.quarkus.test.services.quarkus.ProdQuarkusApplicationManagedResourceBuilder` does inside `io.quarkus.test.services.quarkus.ProdQuarkusApplicationManagedResourceBuilder#findManagedResource` - when `io.quarkus.qe:quarkus-test-openshift` is present, it uses `io.quarkus.test.services.quarkus.OpenShiftQuarkusApplicationManagedResourceBinding` instead of `io.quarkus.test.services.quarkus.ProdLocalhostQuarkusApplicationManagedResource`. You have at least 2 options - introduce another level of abstraction that does same and determines which builder should be used (use delegate pattern), or move `io.quarkus.test.services.quarkus.QuarkusApplicationManagedResourceBuilder#configureCertificates` to `QuarkusApplicationManagedResourceBinding`, by default it does nothing, but for OpenShift it can do something...

michalvavrik:
> It currently supports only 1 certificate for QuarkusApp. Is this a real limitation? I'm not sure I can see a use case where this would be a problem.

I implemented what I needed. I suspect there are some obstacles that would consume time if I didn't limit it. I don't think you really need more.

mocenas:
I've added OCP route and service into cert's subjectAlternateNames for quarkus-build apps. So now TLS-on-OCP should be fully supported for those.

BTW: I rebased this branch on https://github.com/quarkus-qe/quarkus-test-framework/pull/1616 so I can work with SAN injection before the PR is merged. I will rebase it back on main once it is resolved.

mocenas:
run tests

quarkus-test-openshift/src/main/java/io/quarkus/test/services/quarkus/TemplateOpenShiftQuarkusApplicationManagedResource.java
```diff
@@ -71,6 +77,25 @@ private void applyTemplate() {
                 this::internalReplaceDeploymentContent,
                 addExtraTemplateProperties(),
                 model.getContext().getServiceFolder().resolve(DEPLOYMENT));
+
+        if (model.isSslEnabled()) {
+            client.exposeDeploymentPort(model.getContext().getName(), "https", model.getOcpTlsPort());
```
jedla97, line 82: I looked at openshift errors and this part seems to causing it. When I add delay (`Thread.sleep(85);`) between some client call it start working. Also it happening only when running from our openstack machines. So I guess the request are to fast. Will look if I'll find some solution.

jedla97:
run tests

jedla97:
run tests

jedla97:
run tests

jedla97:
Hi @mjurc @fedinskiy the ocp failiure should be fixed and CI is green. I don't know in which state your review finished, if it was ok just need the fix regression or you want to still look at.

The regression was caused by fetching deployment which wasn't fully ready and have populated `resourceVersion`. When the patch was send the old `resourceVersion` clashed with new one. This was fixed by `client.apps().deployments().resource(deployment).unlock().patch();` where `unlock` null the `resourceVersion` so it won't clash.

Other possibility would be create manual wait for getting deployment like `client.apps().deployments().withName(deploymentName).waitUntilCondition()`

mocenas:
Thanks @jedla97 for fixing this. This PR should now be ready. Can you guys review it @fedinskiy @mjurc ?

PR #1651, opened 2025-08-11
Fix running KC in prod mode on Windows

jedla97:
### Summary

Fixing problem caused by propagating KC certs mounting help properties to Quarkus app. This caused the 
```
Running command: java -Dtruststore--1193672267=secret_with_destination::\\quarkus-test-suite\\http\\http-advanced\\target|keycloak-truststore.p12  -jar ..\quarkus-run.jar
13:45:16,458 INFO  [app] 'keycloak-truststore.p12' is not recognized as an internal or external command,
```
Where the `target|keycloak-truststore.p12` causing this error because the `|` is not escaped and windows think there is second command.

Also there was 2nd issue which causing the connection to fail due to not know SAN.  I choose to use `LocalhostManagedResource`  which forwarding localhost to docker host, for that specific port. 

There would be other option to use `DockerClientFactory.instance().dockerHostIpAddress();`. The drawback of it that we would be needed to check if the host is ip or text to use correct prefix `IP:` or `DNS:` per (https://github.com/smallrye/smallrye-certificate-generator/blob/main/certificate-generator-junit5/src/main/java/io/smallrye/certs/junit5/Alias.java#L27)

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakContainerManagedResourceBuilder.java
+++ b/quarkus-test-service-keycloak/src/main/java/io/quarkus/test/services/containers/KeycloakContainerManagedResourceBuilder.java
@@ -14,6 +14,7 @@ import java.util.List;
 import java.util.ServiceLoader;
 import java.util.stream.Collectors;
 
+import io.quarkus.test.bootstrap.LocalhostManagedResource;
 import io.quarkus.test.bootstrap.ManagedResource;
 import io.quarkus.test.bootstrap.OpenShiftExtensionBootstrap;
 import io.quarkus.test.bootstrap.ServiceContext;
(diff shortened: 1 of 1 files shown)


jedla97:
run tests

michalvavrik:
what I mean - this is ok, but the fact that these properties are there is probably not necessary and they are just leaking from somewhere (which in 99 % cases doesn't matter). or am I missing something

jedla97:
> Thanks @jedla97 , on second thought, I think you should had just remote it instead of escaping it 😀 ???

It's happening because of [Certificate#getTrustStoreConfigProperties](https://github.com/quarkus-qe/quarkus-test-framework/blob/main/quarkus-test-core/src/main/java/io/quarkus/test/security/certificate/Certificate.java#L74), it should probably check if the openshift is used as this additional property is only used for mounting. On baremetal it should return only `trust-store`. 

At least I hope we don't and will not need mount anything on baremetal. 

We can probably revert escaping a remove the property by adding openshift/kubernetes check for it. I don't have strong opinion on this. I can do it today @michalvavrik if you agree with other solution.

michalvavrik:
I'd just remove it instead of escaping, we don't need it ever for "java" command. I think there is misunderstanding regarding what we are talking about, these "trust" props are only used by our FW. Yeah, it's a mess, but it is not worth introducing some new concept as we have better things to work on.

Anyway, it is not important, do whatever suits you (if anything).

PR #1655, opened 2025-08-13
Adding not redirect when checking if route is reachable

jedla97:
### Summary

There can be usecases where the base path will be redirected to secured resource. This resource can be secured by self-sign certificate and this check will throw `SSLHandshakeException`.

For example when the OIDC is set as `quarkus.oidc.application-type=web-app` the default behavioral is to redirect to secured resources. The KC in this case is secured by self-sign certificate and this redirection to secured resource cause `SSLHandshakeException`.

If the request to check the route return redirect, it should be enough to know that route is reachable, we don't need to follow it.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:


fedinskiy (approved):
Since it is a bug in RestEasy, there is not much we can do

PR #1669, opened 2025-08-27
Support running Keycloak in production mode on Kubernetes

michalvavrik:
### Summary

Keycloak in production mode uses HTTPS protocol, however our framework only exposes HTTP port, so there is NPE when creating Keycloak `AuthZClient`.

I verified locally this PR fixed ^^^, but my env differs from our daily CI as I am using Minikube with rootless podman and had to do some port forwarding. I do not guarantee that this fixes the test completely in our GH CI.

I have added some logging because when certain exceptions are thrown in JUnit callbacks, it seems to be silent and I didn't know what is happening in there.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-kubernetes/src/main/java/io/quarkus/test/services/containers/KubernetesContainerManagedResource.java
+++ b/quarkus-test-kubernetes/src/main/java/io/quarkus/test/services/containers/KubernetesContainerManagedResource.java
@@ -88,10 +88,14 @@ public class KubernetesContainerManagedResource implements ManagedResource {
         return loggingHandler.logs();
     }
 
-    private void applyDeployment() {
-        String deploymentFile = getConfiguration()
+    protected String getTemplate() {
+        return getConfiguration()
                 .getOrDefault(Configuration.Property.KUBERNETES_DEPLOYMENT_TEMPLATE_PROPERTY,
                         DEPLOYMENT_TEMPLATE_PROPERTY_DEFAULT);
+    }
+
+    private void applyDeployment() {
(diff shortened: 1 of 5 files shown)


jedla97 (approved):
Many thanks! It look good to me. Hopefully it will fix the daily.

rsvoboda:
FYI https://github.com/quarkus-qe/quarkus-test-framework/actions/workflows/daily.yaml failed with various (strange) failures so I retriggered it. There were also outages on Sonatype side.

jedla97:
I see only change what would could  problems in `KeycloakService` but it's only adding try-catch for better propagating exception. Other changes are for Kubernetes only so they should not affect other platform. From my POV that was the reason not to trigger ocp tests

michalvavrik:
@jedla97 @rsvoboda this has no relation to OCP, and I don't believe this could fail daily build. I did test this and even though I did not mention, this will provide more information or fix for all the failures, because remote dev mode also failed on the same status check resulting in the connection refused. I don't see anything wrong about this.

rsvoboda:
Cool, thanks for comments with the details. You managed it well!

michalvavrik:
Thanks, so https://github.com/quarkus-qe/quarkus-test-framework/actions/runs/17288722329/job/49070929037 the remove dev mode is gone (for the moment being), small victory. As for the Keycloak, it fails much further, so what this PR was trying to fix is fixed.

Now I have same failure on my laptop, so I'll try to address it, I just hoped that I was doing the port forwarding wrong and that was the reason.

PR #1689, opened 2025-09-17
Add switch to always include classpath classes

mocenas:
### Summary

Adds feature to include (default) classpath classes, even if custom one are present.

In current behaviour, if any custom class is defined, FW will not include classes on classpath (as it does include them by default). So when I want to add some extra classes, defined in tests but still include (possibly a lot of) classes from src/ I need to manually list them all ([like this](https://github.com/quarkus-qe/quarkus-test-suite/blob/main/hibernate/hibernate-reactive/src/test/java/io/quarkus/ts/hibernate/reactive/MultiDatabaseHibernateReactiveIT.java#L66)). Which is kind of dumb and unnecessary. IMHO there should be a way to easily include default classes in build, even when I add some additional ones.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:


fedinskiy:
@mocenas thank you, I see. In the linked sources you make tests run with classes from `src/main` **and** from `src/test`, and in this PR you want to have an option to always include files from `src/main`. Now this feature at least makes sense :)

On the other hand, I doubt it is a popular pattern. Usually we (or at least me) put all the classes into the `src/main` and pick them with `classes` property only in some very specific cases.

This is why I am not sure, how useful this feature would be; and, as you yourself know, our framework is overcomplicated enough. Therefore, I suggest to first create an issue for this feature and discuss it with a wider team (at least with both Michals).

mocenas:
> This is why I am not sure, how useful this feature would be; and, as you yourself know, our framework is overcomplicated enough. Therefore, I suggest to first create an issue for this feature and discuss it with a wider team (at least with both Michals).

IDK if we need an actual issue for this, IMHO we can discuss here. Opinions on this PR @michalvavrik @mjurc?

michalvavrik:
Approved from my side, but please get broader support for this before you decide to merge it. Ideally find a way to make Fedor happy.

quarkus-test-core/src/main/java/io/quarkus/test/services/QuarkusApplication.java
```diff
@@ -14,6 +14,11 @@
     // By default, it will load all the classes in the classpath.
     Class<?>[] classes() default {};
 
+    /**
+     * always load all the classes in the classpath, even if classes() is set.
+     */
+    boolean alwaysIncludeClasspathClasses() default false;
```
fedinskiy, line 20: Okay, so if we are doing this, please change the name to something more descriptive, since src/test is also on classpath. I suggest something like `includeAllClassesFromMain`, but that is up to you. And make javadoc more detailed, please

fedinskiy (changes requested):
And, please, add some tests, so we can be sure, that this feature works as you intend (downloads all classes from `src/main` and only selected classes from `src/test`)

mjurc (changes requested):
The changes look good, but it should have a docs update + test

mocenas:
I changed the names, add more descriptive comment, changed the actual logic (the were flaws), and added test for it.

fedinskiy (approved):

michalvavrik (approved):
test and docs lgtm, so @mjurc request is handled; I think it is clear enough for me as well;

btw I didn't mention it, but my first thought was that we could use it for Jakarta Data repository, in order to enable these tests, the easiest way is to keep app classes but add PU-specific repositories depending on databases (AKA test class), so I'll use it

PR #1885, opened 2026-04-24
Fix scaleTo() to prevent duplicate pods from DeploymentConfig

jcarranzan:
### Summary

Detected in our Jenkins job: `quarkus-main-rhel8-jdk17-openshift-ts-jvm-ocp-4-stable-weekly/SCENARIO=misc`.
On ocp, a DeploymentConfig may coexist with Deployment created (by OpenShift extension). The current `scaleTo()` only targets the Deployment, leaving DC pods active. So this results in duplicate pods per service with inconsistent app state , for instance, in the failed job, split atomicInteger counters across JVMs : 
```
[ERROR] io.quarkus.ts.infinispan.client.OperatorOpenShiftInfinispanCountersIT.testMultipleClientIncrement -- Time elapsed: 1.814 s <<< FAILURE!
03:47:04 org.opentest4j.AssertionFailedError: expected: <10> but was: <4>
```

With this fix scales down the DC to 0 if one exists, if not DC is present then nothing happens.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
+++ b/quarkus-test-openshift/src/main/java/io/quarkus/test/bootstrap/inject/OpenShiftClient.java
@@ -443,7 +443,9 @@ public final class OpenShiftClient {
     }
 
     /**
-     * Scale the service to the replicas.
+     * Scale the service to the given number of replicas.
+     * Also scales down any existing DeploymentConfig to 0, since on some OCP clusters a DC may
+     * coexist with the Deployment created by the Quarkus OpenShift extension, causing duplicate pods.
      *
      * @param service
      * @param replicas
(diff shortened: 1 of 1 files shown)


jcarranzan:
run tests

mocenas:
I don't like this thing, we can use it as a workaround but it is not solving the problem.
The main problem IMHO is - where did we get a deploymentConfig from? That thing is deprecated for years and should not be used.
Also how is it possible that we got both - deployment and deploymentConfig? If we got both of them there was something terribly wrong.

jcarranzan:
> I don't like this thing, we can use it as a workaround but it is not solving the problem. The main problem IMHO is - where did we get a deploymentConfig from? That thing is deprecated for years and should not be used. Also how is it possible that we got both - deployment and deploymentConfig? If we got both of them there was something terribly wrong.

Yeah, good point, let me investigate it deeper and let you know my insights later, thanks @mocenas

mjurc (changes requested):
What Martin said, let's have solution at the root cause of the issue

jedla97:
Afaik the `OperatorOpenShiftInfinispanCountersIT` don't fail with 3.33. Couldn't be possible that it's bug in Quarkus?

jcarranzan:
> Afaik the `OperatorOpenShiftInfinispanCountersIT` don't fail with 3.33. Couldn't be possible that it's bug in Quarkus?

Yes it's quite probably and it seems the root issue is caused here --> https://github.com/quarkusio/quarkus/pull/53421 
Investigating...

jcarranzan:
I've created an upstream issue [53797](https://github.com/quarkusio/quarkus/issues/53797 )

jcarranzan:
The issue was already resolved, they created upstream fix here https://github.com/quarkusio/quarkus/pull/53778  and merged into main. Jenkins job is now green. Closing it

PR #1886, opened 2026-04-24
Use custom build of Quarkus in CI

fedinskiy:
### Summary

This will make sure, that we use platform build from CI an not the upstrem one

part of https://redhat.atlassian.net/browse/QQE-2465

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,7 @@
             <url>https://central.sonatype.com/repository/maven-snapshots/</url>
             <snapshots>
                 <enabled>true</enabled>
+                <updatePolicy>never</updatePolicy>
             </snapshots>
             <releases>
                 <enabled>false</enabled>
@@ -94,12 +95,13 @@
         <pluginRepository>
             <id>quarkus-snapshots-plugin-repository</id>
             <url>https://central.sonatype.com/repository/maven-snapshots/</url>
+            <snapshots>
+                <enabled>true</enabled>
+                <updatePolicy>never</updatePolicy>
+            </snapshots>
             <releases>
                 <enabled>false</enabled>
             </releases>
-            <snapshots>
-                <enabled>true</enabled>
-            </snapshots>
         </pluginRepository>
     </pluginRepositories>
     <dependencyManagement>


fedinskiy:
In draft state, since I am running local checks and we will need changes on Jenkins side for ocp tests.

mjurc (changes requested):
I don't think this is the right solution if you really want the maven run not to use remote in accordance with update policy, the real fix would be to use `-nsu` or `--no-snapshot-updates` in the run of test suite

But I'm not really sure what issue this is supposed to solve

.github/workflows/build-quarkus.yml
```diff
@@ -26,13 +26,19 @@ jobs:
...
           echo "LOCAL_REPO=$(pwd)/maven-repository" >> $GITHUB_ENV
-      - name: Build Quarkus 999-SNAPSHOT
+      - name: Build Quarkus 999-SNAPSHOT+qe
         run: |
-          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Dno-test-modules -Prelocations -Dmaven.repo.local=${{ env.LOCAL_REPO }}
-      - name: Build Quarkus Platform 999-SNAPSHOT
+          git clone https://github.com/quarkusio/quarkus.git && cd quarkus
+          # We need to make sure, that tests use this build (which includes platform) and not download the upstream one.
+          # To distinguish them we use build metadata suffix:https://semver.org/#spec-item-10
+          grep --exclude-dir=.git -rlw '999-SNAPSHOT' | xargs -d'\n' sed -i -e 's/999-SNAPSHOT/999-SNAPSHOT+qe/g'
```
jedla97: If I understand this correctly this changing the version in project. Wouldn't be better to use `./mvnw versions:set -DnewVersion=999-SNAPSHOT+qe` instead of bash?

I have something like this for branch build of platform (https://github.com/quarkus-qe/quarkus-test-framework/blob/1.8.z/.github/workflows/pr.yaml#L47)

mjurc: I also don't much see a point in chaning the version of `io.quarkus:quarkus-bom` version, it is a snapshot build and if we don't want to use what's in remote we should use `-nsu` in execution where we don't want it

.github/workflows/daily-deploy.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/deploy
         with:
-          release-version: '999-SNAPSHOT'
+          release-version: '999-SNAPSHOT+qe'
```
mjurc: I don't think we want to change this

mjurc (changes requested):

mjurc:
A few more items:
- we only use sonatype snapshot central to get Quarkus. If we don't want to use it and we cannot depend on getting proper snapshots from it, and we want to use bits all built in our CI, maybe it would be worth considering removing the repository from the pom and build the snapshots ourselves without weird version numbers
- that way we can keep pushing framework nightly into snapshot repo without weird version numbers too :)

.github/workflows/build-quarkus.yml
fedinskiy: This could be done for platform, but fails for quarkus, since it also includes gradle modules, which are not updated

fedinskiy:
@mjurc this is being done to not mix with version uploaded to sonatype. Was requested by @rsvoboda in our slack

rsvoboda:
`-nsu` sounds like a good approach for GH actions where we unpack maven repo

Version change felt like a good thingy, so that's why I proposed that. But looking at the number of changes I don't think this is a good path.

The main goal is to avoid interference with snapshots of platform deployed from https://github.com/quarkusio/quarkus-platform/blob/main/.github/workflows/deploy-snapshots.yml
The goal is to find simple and elegant solution which doesn't bite us elsewhere.

So maybe dropping sonatype snapshots and building also FW snapshots and later using `-nsu` would be better.
On Jenkins side I think we could get rid of sonanatype snapshots mirror and deploy FW on our own together with quarkus core and platform.

jedla97:
Afaik we define the sonatype repository here https://github.com/quarkus-qe/quarkus-test-framework/blob/main/pom.xml#L80

By default it's forcing the update daily (I spot it now in TS PR). I think that we should set the `<updatePolicy>never</updatePolicy>` and it would have similar effect as `-nsu`

fedinskiy:
Applied Jakub's suggestion in regards to disabling the updates. I prefer it to adding `-nsu`, since the latter requires updating all the same places as build prefix solution.

jedla97 (approved):
I thinks this should be OK and have the expected effect, base on docs description https://maven.apache.org/settings.html#repositories

Probably something similar should be done for TS, but there we want probably update it for local env, so when people run test locally they will have latest FW available

mjurc (approved):
LGTM, thank you

PR #1897, opened 2026-05-19
Adding DisabledOnFips annotation

jedla97:
### Summary

This PR adding the `DisabledOnFips` annotation. We currently have only `DisabledOnFipsAndNative`.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/DisabledOnFipsCondition.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/DisabledOnFipsCondition.java
@@ -1,15 +1,12 @@
 package io.quarkus.test.scenarios.annotations;
 
+import static io.quarkus.test.utils.FipsUtils.isFipsEnabledEnvironment;
+
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
 public class DisabledOnFipsCondition implements ExecutionCondition {
-
-    /**
-     * We set environment variable "FIPS" to "fips" in our Jenkins jobs when FIPS are enabled.
-     */
-    private static final String FIPS_ENABLED = "fips";
 
     @Override
     public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
@@ -18,10 +15,6 @@ public class DisabledOnFipsCondition implements ExecutionCondition {
         }
 
         return ConditionEvaluationResult.enabled("The test is not running in FIPS enabled environment");
-    }
-
-    private static boolean isFipsEnabledEnvironment() {
-        return FIPS_ENABLED.equalsIgnoreCase(System.getenv().get("FIPS"));
     }
 
 }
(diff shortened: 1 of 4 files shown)


quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/DisabledOnFipsCondition.java
```diff
@@ -0,0 +1,27 @@
...
+    public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+        if (isFipsEnabledEnvironment()) {
+            return ConditionEvaluationResult.disabled("The test is running in FIPS enabled environment");
+        }
+
+        return ConditionEvaluationResult.enabled("The test is not running in FIPS enabled environment");
+    }
+
+    private static boolean isFipsEnabledEnvironment() {
+        return FIPS_ENABLED.equalsIgnoreCase(System.getenv().get("FIPS"));
```
fedinskiy: Please, reuse the method from DisabledOnFipsAndNativeCondition. Or reuse method from this class there. Or move this method to a separate utility class. We need a single source of truth for this

fedinskiy (changes requested):
In addition to the above, doesn't test failure on FIPS mean there is a bug, that should be fixed ASAP? In that case even simple `@Disabled` should suffice. Do we really need that level of granularity for disabled tests?

jedla97:
> In addition to the above, doesn't test failure on FIPS mean there is a bug, that should be fixed ASAP? In that case even simple `@Disabled` should suffice. Do we really need that level of granularity for disabled tests?

When I investigated one of the failure this appear https://redhat.atlassian.net/browse/QUARKUS-7816

Basically it's dev mode on FIPS where default MariaDB don't start. If we use `@Disable` we don't have coverage for this case on other platforms/enviroments. It's kinda related to https://github.com/quarkus-qe/quarkus-test-suite/issues/2431 which I fixed on side of our testsuite. In my opinion this one won't be probably fixed as in the end user can overwrite the default database, which we doing in other devmode test for MariaDB. Also I don't believe that there are users which doing development on FIPS system and we maybe don't even support dev mode on FIPS, but there I'm not sure.

quarkus-test-core/src/main/java/io/quarkus/test/scenarios/annotations/DisabledOnFipsCondition.java
jedla97: Updated

quarkus-test-core/src/main/java/io/quarkus/test/utils/FipsUtils.java
```diff
@@ -0,0 +1,17 @@
...
+     * We set environment variable "FIPS" to "fips" in our Jenkins jobs when FIPS are enabled.
+     */
+    private static final String FIPS_ENABLED = "fips";
+
+    private FipsUtils() {
+
+    }
+
+    public static boolean isFipsEnabledEnvironment() {
+        return FIPS_ENABLED.equalsIgnoreCase(System.getenv().get("FIPS"));
```
gastaldi, line 15: So the environment should have `export FIPS=fips`? that's kinda odd :)

gastaldi: Ah I see this is already used in the pom.xml: https://github.com/quarkus-qe/quarkus-test-suite/blob/main/pom.xml#L1054-L1055

gastaldi (approved):

fedinskiy (approved):
Thank you!

quarkus-test-core/src/main/java/io/quarkus/test/utils/FipsUtils.java
jedla97: > So the environment should have export FIPS=fips?

We setting it on our RHEL FIPS runners, it was probably done as most simple solution at that time to use it in code + pom.xml.

PR #1910, opened 2026-05-29
Use temp directory by default and init git repo in gitignored dirs

gastaldi:
## Summary
- The `rewrite-maven-plugin` 6.40.0 (shipped in `quarkus-updates` 1.11.0) fixed its gitignore handling so files in gitignored+untracked directories are correctly excluded from OpenRewrite processing
- Since test apps were created inside `target/` (which is gitignored by the project), OpenRewrite skipped their `pom.xml` and no dependency renames were applied, causing `QuarkusCliUtilsExampleIT.exampleDependencyUpdateTest` to fail
- Two complementary fixes:
  - **Default to temp directory** when no explicit `targetFolderName` is specified in `createApplication()` and `createApplicationFromExistingSources()`, so apps are created outside the git tree
  - **Initialize a git repo** after app creation when the app directory is inside a gitignored path (e.g. via `createApplicationAt()`), scoping OpenRewrite to the app's own directory

Fixes: [QQE-2579](https://redhat.atlassian.net/browse/QQE-2579)

## Test plan
- [x] Verified `QuarkusCliUtilsExampleIT.exampleDependencyUpdateTest` fails without the fix
- [x] Verified the test passes with the fix applied

### DIFF
--- a/quarkus-test-cli/src/main/java/io/quarkus/test/bootstrap/QuarkusCliClient.java
+++ b/quarkus-test-cli/src/main/java/io/quarkus/test/bootstrap/QuarkusCliClient.java
@@ -5,7 +5,6 @@ import static io.quarkus.test.services.quarkus.model.QuarkusProperties.createDis
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
-import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -46,7 +45,6 @@ public class QuarkusCliClient {
             + " to be unavailable";
     private static final PropertyLookup COMMAND = new PropertyLookup(CLI_CMD.getName(), "quarkus");
     private static final Path TARGET = Paths.get("target");
-    private volatile boolean useTemporaryDirectory = false;
 
     private final ScenarioContext context;
 
@@ -106,17 +104,16 @@ public class QuarkusCliClient {
     }
 
     public QuarkusCliRestService createApplication(String name, CreateApplicationRequest request, String targetFolderName) {
+        String dirName = extractDirectoryName(name);
         final Path serviceFolder;
         if (isNotEmpty(targetFolderName)) {
-            serviceFolder = TARGET.resolve(targetFolderName).resolve(name);
-        } else if (useTemporaryDirectory) {
+            serviceFolder = TARGET.resolve(targetFolderName).resolve(dirName);
+        } else {
             try {
-                serviceFolder = Files.createTempDirectory(name).resolve(name);
+                serviceFolder = Files.createTempDirectory(dirName).resolve(dirName);
             } catch (IOException e) {
                 throw new RuntimeException("Failed to create temporary directory", e);
             }
-        } else {
-            serviceFolder = null;
         }
         QuarkusCliRestService service = new QuarkusCliRestService(this, serviceFolder);
         ServiceContext serviceContext = service.register(name, context);
@@ -149,6 +146,8 @@ public class QuarkusCliClient {
         abortIfRegistryUnavailable(result, "application");
         assertTrue(result.isSuccessful(), "The application was not created. Output: " + result.getOutput());
 
+        initGitRepoIfInsideGitIgnoredDir(serviceContext.getServiceFolder());
+
         return service;
     }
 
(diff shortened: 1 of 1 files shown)


rsvoboda:
Can we assume `git` command always available?

mocenas:
> Can we assume `git` command always available?

I think we have `git` installed everywhere.

mocenas:
I think this is OK and it solves the issue. But maybe init the git when we're creating the app instead of updating it? (seems more logical). I'm not sure if we're updating the app twice sometimes or if there are any other possible issues.

gastaldi:
Moved `git init` to app creation phase

mocenas:
The changes are OK. One more thing, please add there a comment explaining why the git init is necessary. I can imagine that somebody without knowledge of this issue will be very very confused when reading it.

When you think about it from outside point of view, it seemingly doesn't make sense to init git (which is then never used) in an app which is temporary and will be removed after tests are done.

rsvoboda:
@jedla97 suggested that `cliClient.useTemporaryDirectory()` can be used

I will review this PR and try the suggestion later today

jedla97:
The tmp dir was introduced here https://github.com/quarkus-qe/quarkus-test-framework/pull/1729 

I believe it can be set to use tmp dir by default and either remove the usage of local dir if the `targetFolderName` is null or rename the `useTemporaryDirectory` -> `useTestTargetDirectiry` or something like this.


Also note if you go with tmp dir or git init, this will need to be backported to 1.8.z and 1.7.z FW branches and FW released otherwise the tests will fail as the `quarkus-update` is not based on branches and used latest "released" version

rsvoboda:
ok, so we can use "git init" approach or "useTemporaryDirectory()".

I'm inclining slightly more towards "useTemporaryDirectory()" (like 51/49) approach, but I would like to hear more from @gastaldi @jedla97 @mocenas here

```diff
diff --git a/quarkus-test-cli/src/main/java/io/quarkus/test/util/DefaultQuarkusCLIAppManager.java b/quarkus-test-cli/src/main/java/io/quarkus/test/util/DefaultQuarkusCLIAppManager.java
index 10420495..a923ecf1 100644
--- a/quarkus-test-cli/src/main/java/io/quarkus/test/util/DefaultQuarkusCLIAppManager.java
+++ b/quarkus-test-cli/src/main/java/io/quarkus/test/util/DefaultQuarkusCLIAppManager.java
@@ -33,6 +33,7 @@ public class DefaultQuarkusCLIAppManager implements IQuarkusCLIAppManager {
     @Override
     public QuarkusCliRestService createApplication() {
         Log.info("Creating app with version stream: " + oldVersion);
+        cliClient.useTemporaryDirectory();
         return cliClient.createApplication("app", defaults()
                 .withPlatformBom(null)
                 .withStream(oldVersion.toString())
```

gastaldi:
I don't have a strong opinion here — whichever the team prefers is fine

mocenas:
I also prefer the temporary directory more.

rsvoboda (approved):
Let's have the fix in QuarkusCliClient so we do not need to touch the tests.

rsvoboda:
> I also prefer the temporary directory more.

:) I'm so 50/50

rsvoboda:
Let's hear from Jakub, both approaches have some impact ...

gastaldi:
> The changes are OK. One more thing, please add there a comment explaining why the git init is necessary. I can imagine that somebody without knowledge of this issue will be very very confused when reading it.

Done

jedla97:
I'm prefer the tmp dir slightly more as well as we have using it in some update tests already.

gastaldi:
The safest approach is put the git init in updateApplication() as a defensive measure. That way it covers all cases — whether the app is in target/, a temp dir, or anywhere else under a parent git repo with a gitignore. The temp dir change remains the primary fix, and git init before update is the safety net.

gastaldi:
Two changes were applied:

  1. Temp dir by default — apps without explicit targetFolderName go to a temp directory (both create methods)
  2. Conditional git init at creation — when the app is in a gitignored dir (e.g. via createApplicationAt), initialize a git repo right after creation so any subsequent CLI operation works correctly

quarkus-test-cli/src/main/java/io/quarkus/test/bootstrap/QuarkusCliClient.java
```diff
@@ -109,14 +109,12 @@ public QuarkusCliRestService createApplication(String name, CreateApplicationReq
         final Path serviceFolder;
         if (isNotEmpty(targetFolderName)) {
             serviceFolder = TARGET.resolve(targetFolderName).resolve(name);
-        } else if (useTemporaryDirectory) {
```
rsvoboda, line 112: `useTemporaryDirectory` method and field should be dropped

mocenas (commented):
I agree with @rsvoboda we should remove the `useTemporaryDirectory` field and method, since it is no longer doing anything. Otherwise good.

rsvoboda:
CLI examples fail

gastaldi:
@rsvoboda it's unrelated to this change. Like https://redhat.atlassian.net/browse/QQE-2587 it's Quay.io being unstable

gastaldi:
Failures are unrelated to this change: 
1. shouldCreateApplicationUsingArtifactId — expected: `<my-app>` but was: `<com.mycompany:my-app>` — this is a Quarkus CLI behavior change in how it names directories, not related to our fix
2. RemoteDevModeGreetingResourceIT — 1 error, 0 failures — likely the same flaky timeout seen before

PR #1927, opened 2026-06-19
Trigger daily deploy snapshot on push to main

gastaldi:
## Summary
- Add `push` trigger on `main` branch to the daily deploy workflow so snapshots are deployed on every push, not just on the nightly schedule.

## Test plan
- [ ] Verify the workflow triggers on push to `main`
- [ ] Verify the scheduled and manual dispatch triggers still work

### DIFF
--- a/.github/workflows/daily-deploy.yml
+++ b/.github/workflows/daily-deploy.yml
@@ -1,5 +1,8 @@
 name: "Daily deploy snapshot"
 on:
+  push:
+    branches:
+      - main
   workflow_dispatch:
   schedule:
     - cron: '0 23 * * *'


jedla97 (commented):
TBH I don't think if it's the good idea especially as the sonatype pushing for more limits. In general if the person want to wait for daily deploy, they can trigger the build.

fedinskiy (commented):
I like the idea, but it looks like we need to have a wider discussion :)

@gastaldi could you raise this in our slack or during the call?

rsvoboda:
I don't think it's necessary, there is a daily https://github.com/quarkus-qe/quarkus-test-framework/actions/workflows/daily-deploy.yml which can also be triggered manually.

gastaldi:
Thanks for the feedback. Now that I think of it, not every push may require publishing (eg. If tests are changed). Closing

PR #1944, opened 2026-07-09
Explicitly show, that there is no need to update the test file

fedinskiy:
### Summary
This file was added for tests only (see its path) so it can use any version and doesn't need to be updated

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-core/src/test/resources/quarkus-app-pom.xml
+++ b/quarkus-test-core/src/test/resources/quarkus-app-pom.xml
@@ -6,9 +6,10 @@
   <artifactId>examples-external-applications</artifactId>
   <version>999-SNAPSHOT</version>
   <properties>
+    <!--  This file is only used by QuarkusMavenPluginBuildHelperTest, so no need to update values below -->
     <maven.compiler.release>17</maven.compiler.release>
     <surefire-plugin.version>3.5.4</surefire-plugin.version>
-    <quarkus.platform.version>3.999-SNAPSHOT</quarkus.platform.version>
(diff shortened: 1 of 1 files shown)


quarkus-test-core/src/test/resources/quarkus-app-pom.xml
```diff
@@ -8,7 +8,7 @@
   <properties>
     <maven.compiler.release>17</maven.compiler.release>
     <surefire-plugin.version>3.5.4</surefire-plugin.version>
-    <quarkus.platform.version>3.999-SNAPSHOT</quarkus.platform.version>
+    <quarkus.platform.version>WHATEVER</quarkus.platform.version>
```
jedla97: So If I understand correctly it's here mainly because of use later in this pom file and it's overwritten every time.

If that is so, could you maybe add comment to it so it's clear to whoever will be reading this.

fedinskiy: It is here to keep the normal file structure, but this file is only needed for QuarkusMavenPluginBuildHelperTest class and we do not check this line there

jedla97: Thanks, I would still prefer some comments there so people don't need to check this PR for explanation

fedinskiy: This is also written in the commit description and can be reasoned based on folder path. Added a comment for even more clarity.

jedla97 (approved):
Thanks

PR #1963, opened 2026-08-10
Support authenticated Maven repositories for OCP S2I source builds

gtroitsk:
### Summary

Add support for separate releases/snapshots Maven repositories with optional authentication for OCP S2I source builds (e.g. TaaS Artifactory).

New configuration properties:
- `ts.global.s2i.maven.releases.repository`  - URL for releases repo
- `ts.global.s2i.maven.snapshots.repository` - URL for snapshots repo
- `ts.global.s2i.maven.remote.repository.username` - repo server username
- `ts.global.s2i.maven.remote.repository.password` - repo server password

When releases/snapshots repos are configured, `settings-mvn-auth.yml` is used instead of setting-mvn.yml. All placeholders (repo URLs and credentials) are resolved in Java before the ConfigMap is uploaded to
the cluster, so credentials are present in settings.xml from the start of the S2I build.

The existing `s2i.maven.remote.repository` (single anonymous URL) option is
fully preserved and unchanged.

`settings-mvn-auth.yml` is a new Maven settings template used for the
authenticated split-URL case; `settings-mvn.yml `is left unchanged.

- [ ] Dependency update
- [ ] Refactoring

### Checklist:

### DIFF
--- a/quarkus-test-openshift/src/main/resources/settings-mvn-auth.yml
+++ b/quarkus-test-openshift/src/main/resources/settings-mvn-auth.yml
@@ -1,0 +1,49 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: settings-mvn
+data:
+  settings.xml: |
+    <?xml version="1.0" encoding="UTF-8"?>
+    <settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
+        <servers>
+            ${servers.block}
+        </servers>
+        <mirrors>
+            ${central.mirror.block}
+        </mirrors>
+        <profiles>
+            <profile>
+                <id>redhat</id>
+                <repositories>
+                    <repository>
+                        <id>redhat</id>
+                        <name>Repository for Redhat dependencies</name>
+                        <url>https://maven.repository.redhat.com/ga/</url>
+                    </repository>
+                </repositories>
+                <pluginRepositories>
+                    <pluginRepository>
+                        <id>redhat</id>
+                        <name>Repository for Redhat dependencies</name>
+                        <url>https://maven.repository.redhat.com/ga/</url>
+                    </pluginRepository>
+                </pluginRepositories>
+            </profile>
+            <profile>
+                <id>customRepository</id>
+                <repositories>
+                    ${snapshots.repository.block}
+                </repositories>
+                <pluginRepositories>
+                    ${snapshots.pluginRepository.block}
+                </pluginRepositories>
+            </profile>
+        </profiles>
+        <activeProfiles>
+            <activeProfile>customRepository</activeProfile>
+            <activeProfile>redhat</activeProfile>
+        </activeProfiles>
+    </settings>
--- a/quarkus-test-core/src/main/java/io/quarkus/test/configuration/Configuration.java
+++ b/quarkus-test-core/src/main/java/io/quarkus/test/configuration/Configuration.java
@@ -204,6 +204,10 @@ public final class Configuration {
         CONTAINER_REGISTRY_URL("container.registry-url"),
         CONTAINER_PREFIX("docker-container-prefix"),
         S2I_MAVEN_REMOTE_REPOSITORY("s2i.maven.remote.repository"),
+        S2I_MAVEN_RELEASES_REPOSITORY("s2i.maven.releases.repository"),
+        S2I_MAVEN_SNAPSHOTS_REPOSITORY("s2i.maven.snapshots.repository"),
+        S2I_MAVEN_REMOTE_REPOSITORY_USERNAME("s2i.maven.remote.repository.username"),
+        S2I_MAVEN_REMOTE_REPOSITORY_PASSWORD("s2i.maven.remote.repository.password"),
         S2I_REPLACE_CA_CERTS("s2i.java.replace-ca-certs"),
         S2I_BASE_NATIVE_IMAGE("s2i.openshift.base-native-image"),
         CUSTOM_BUILD_REQUIRED("custom-build.required");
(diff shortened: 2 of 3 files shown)


mjurc (commented):
I think this looks quite well but could you please fix imports so the CI can proceed?

gtroitsk:
> I think this looks quite well but could you please fix imports so the CI can proceed?

I will update the PR when I have a working solution

gtroitsk:
I want to run `external-applications`  tests with TaaS artifactory snapshot repo on Jenkins but there are no credentials at this moment.  I will prepare follow up MR later today

mjurc:
run kubernetes && run tests

mjurc:
@gtroitsk alright, I'd like to see that too so please share the job with me in DM

mjurc:
run tests

quarkus-test-openshift/src/main/resources/settings-mvn-auth.yml
```diff
@@ -0,0 +1,66 @@
...
+            </server>
+            <server>
+                <id>internal.s2i.maven.snapshots.repository</id>
+                <username>${maven.repo.username}</username>
+                <password>${maven.repo.password}</password>
+            </server>
+        </servers>
+        <mirrors>
+            <mirror>
+                <id>google-maven-central</id>
```
mjurc: The releases repository configured by `s2i.maven.releases.repository` should be the mirror of central. I wouldn't like to fallback onto another mirror that's going to rate limit us when we use our mirrors for everything else, one that's not configurable.

I guess `https://maven-central.storage-download.googleapis.com/maven2/` could be default value for  `s2i.maven.releases.repository` if it is unset though

quarkus-test-openshift/src/main/resources/settings-mvn-auth.yml
```diff
@@ -0,0 +1,66 @@
...
+                <mirrorOf>central</mirrorOf>
+                <name>Google Maven Central mirror</name>
+                <url>https://maven-central.storage-download.googleapis.com/maven2/</url>
+                <blocked>false</blocked>
+            </mirror>
+        </mirrors>
```
mjurc: We also configure sonatype central snapshots repo in test suite. the repository configured by `ts.global.s2i.maven.snapshots.repository` should be a mirror of that.

mjurc (commented):
Besides the note, is there any reason to keep the old configuration? These settings do everything the old config did, but they also allow to configure authenticated access and snapshots repository.

quarkus-test-openshift/src/main/resources/settings-mvn-auth.yml
gtroitsk: I have a few clarifying questions before I update the PR:

1. When `s2i.maven.releases.repository` is set, should it act as a mirror for Maven Central?
2. When `s2i.maven.snapshots.repository` is set, the Google mirror might also cause issues. Should we ensure our Central mirror repository is used instead?
3. Do we want to update all jobs to use the TaaS Artifactory?
4. I plan to remove the old property in a separate PR once the pipelines are updated to use the new config

mjurc: > When s2i.maven.releases.repository is set, should it act as a mirror for Maven Central?

Yes. If it is not set, imo it should set `https://maven-central.storage-download.googleapis.com/maven2/`. It should also be only used for releases.

> When s2i.maven.snapshots.repository is set, the Google mirror might also cause issues. Should we ensure our Central mirror repository is used instead?

It shouldn't cause issues if the first repository only mirrors `central` since that's release only repository. I don't have a good answer here - the default Maven config does not configure any snapshot repository, so if no `s2i.maven.snapshots.repository` is set, it should just not be added to the generated config

> Do we want to update all jobs to use the TaaS Artifactory?

All the jobs on new infra have to use that as they don't have access to legacy mirror. In the old infra, the ones that don't have access to legacy mirror (aarch64 clusters) should be updated

> I plan to remove the old property in a separate PR once the pipelines are updated to use the new config

sounds good to me

gtroitsk: Updated the default release repo to the Google mirror with an option to use a custom repo instead. It makes sense to run combinations with and without auth and with released and snapshot Quarkus versions. However, this is currently blocked until the Quarkus deployment is fixed because 3.x platform artifacts are no longer in Artifactory.

llowinge:
I want to ask, if scenario where i will use `s2i.maven.remote.repository` which will point to our CQ QE Artifactory (+ setting the auth via `s2i.maven.remote.repository.username/password` as it needs auth) will still work? As i think i don't see `s2i.maven.remote.repository` being used in the `settings-mvn-auth.xml`?

Or should i swap `s2i.maven.remote.repository` -> `s2i.maven.releases.repository` + username/password ?

gtroitsk:
> I want to ask, if scenario where i will use `s2i.maven.remote.repository` which will point to our CQ QE Artifactory (+ setting the auth via `s2i.maven.remote.repository.username/password` as it needs auth) will still work? As i think i don't see `s2i.maven.remote.repository` being used in the `settings-mvn-auth.xml`?
> 
> Or should i swap `s2i.maven.remote.repository` -> `s2i.maven.releases.repository` + username/password ?

Please switch to `s2i.maven.releases.repository` if you require authentication. `s2i.maven.remote.repository` is deprecated and will be removed once all jobs have been updated to use new properties.

mjurc:
@gtroitsk can I ask you to test this solution also with an unauthenticated repository too? I wouldn't like to cut those off completely, I'd just like to have the ability to specify authentication if I need it for the repo

gtroitsk:
> @gtroitsk can I ask you to test this solution also with an unauthenticated repository too? I wouldn't like to cut those off completely, I'd just like to have the ability to specify authentication if I need it for the repo

Yes, that should work for both variants. When we have the snapshot artifacts in Artifactory, I will share a Jenkins job that runs a set of external applications with property combinations that make sense.

mjurc (approved):
@gtroitsk can you please open backports into live streams asap?

Also LGTM, thanks!

